PR #36, opened 2020-12-20
Fix ForSupport import in documentation

spinscale:
The `ForSupport` class is in another package in 1.5.0


casid:
Aww, thanks for spotting this!

Issue #16, opened 2020-09-26
Gradle plugin feature request


mtdowling:
This project looks great! I just wanted to submit a feature request for a Gradle plugin.

casid:
Hey @mtdowling, thanks for the heads up! I'm not very experienced with gradle (except for Android development). Would you be interested in helping with a Gradle plugin?

casid:
I just had a quick look how this could be done. I think it would be great if we could add the gradle plugin as a module to this repository and build it together with the rest of jte. This way we can make sure that the gradle plugin is always up-to-date and works with the current jte version.

I'd suggest to put it into a `jte-gradle-plugin` module. The main build runs with maven though. It looks like the Spring Boot team faced the same problem and was able to build their gradle plugin from maven.

[This stackoverflow post](https://stackoverflow.com/questions/27555560/how-can-i-build-a-gradle-plugin-with-maven) looks like a good starting point I guess.

Jerbell:
This is not a proper solution, but as an interim measure you can use a `JavaExec` task rather than a proper Gradle task or plugin.  This uses Shadow Jar Gradle plugin and Kotlin, but could be adjusted to work with Java.

In `build.gradle.kts`:
```kotlin
val precompileJteTemplates = tasks.register<JavaExec>("precompileJteTemplates") {
    description = "Precompile all JTE templates files."
    group = "build"
    main = "com.flowcrypt.utils.JteTemplatePrecompiler"
    val mainSourceSet = project.the<SourceSetContainer>()["main"]
    classpath = mainSourceSet.compileClasspath.plus(mainSourceSet.output)
    shouldRunAfter("compileKotlin")
    outputs.upToDateWhen { false } // not sure if this is necessary
}

tasks.named("shadowJar") {
    dependsOn(precompileJteTemplates)
}
```

And the class itself (uses Kotlin logging):
```kotlin
package com.flowcrypt.utils

import gg.jte.ContentType
import gg.jte.TemplateEngine
import gg.jte.resolve.DirectoryCodeResolver
import mu.KotlinLogging
import java.nio.file.Path
import java.util.concurrent.TimeUnit

object JteTemplatePrecompiler {

    private val logger = KotlinLogging.logger {}

    // these ideally shouldn't be hard coded
    private val sourceDirectory: String = "src/main/jte"
    private val targetDirectory: String = "build/resources/main"
    private val compilePath: List<String> = emptyList() // redundant
    private val contentType: String = "Html"

    @JvmStatic
[...]

PR #53, opened 2021-02-12
Add gradle plugin

iRebbok:
This PR adds a gradle plugin with two tasks, `CompilerMojo` & `GeneratorMojo`. I'm not a pro at gradle, so something might be of low quality.

Related issue #16.

<details>
<summary>To use the tasks locally</summary>

Put that into `settings.gradle.build`:
```kts
pluginManagement {
    buildscript {
        repositories {
            flatDir {
                dir("PATH TO A FOLDER WITH A BUILT PLUGIN")
            }
        }

        dependencies {
            classpath("gg.jte:jte-gradle-plugin:1.6.1-SNAPSHOT") // Remember to keep the version updated
        }
    }
}
```

And from `build.gradle.kts` you can use the tasks:
```kts
import java.nio.file.Path
import gg.jte.ContentType

plugins {
    id("gg.jte.gradle")
}

val jtePath = Path.of(project.projectDir.absolutePath,"src", "jte")

tasks.generatorMojo {
    sourceDirectory = jtePath
    contentType = ContentType.Html
}

tasks.compilerMojo {
    sourceDirectory = jtePath
    targetDirectory = Path.of(project.projectDir.absolutePath, "artifacts", "jte-compiled")
    contentType = ContentType.Html
}

tasks.build {
    dependsOn(tasks.compilerMojo)
}
```

</details>

### DIFF
--- a/jte-gradle-plugin/src/main/java/gg/jte/gradle/JteGradle.java
+++ b/jte-gradle-plugin/src/main/java/gg/jte/gradle/JteGradle.java
@@ -1,0 +1,12 @@
+package gg.jte.gradle;
+
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+
+public class JteGradle implements Plugin<Project> {
+    @Override
+    public void apply(Project project) {
+        project.getTasks().register("generatorMojo", GeneratorMojoTask.class);
+        project.getTasks().register("compilerMojo", CompilerMojoTask.class);
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -47,11 +47,11 @@
         <module>jte-runtime</module>
         <module>jte</module>
         <module>jte-maven-plugin</module>
+        <module>jte-gradle-plugin</module>
         <module>jte-runtime-test</module>
         <module>jte-runtime-cp-test</module>
         <module>jte-test-report</module>
         <module>jte-deploy-nexus</module>
-        <module>jte-gradle-plugin</module>
     </modules>
 
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,7 @@
         <module>jte-runtime-cp-test</module>
         <module>jte-test-report</module>
         <module>jte-deploy-nexus</module>
+        <module>jte-gradle-plugin</module>
     </modules>
 
     <dependencies>
--- a/jte-gradle-plugin/pack-gradle-plugin.xml
+++ b/jte-gradle-plugin/pack-gradle-plugin.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<assembly>
+    <id>pack-gradle-plugin</id>
+    <formats>
+        <format>jar</format>
+    </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
+    <dependencySets>
+        <dependencySet>
+            <outputDirectory>.</outputDirectory>
+            <useProjectArtifact>false</useProjectArtifact>
+            <unpack>true</unpack>
+            <scope>runtime</scope>
+            <includes>
+                <include>gg.jte:jte</include>
+                <include>gg.jte:jte-runtime</include>
+            </includes>
+        </dependencySet>
+    </dependencySets>
+    <fileSets>
+        <fileSet>
+            <outputDirectory>.</outputDirectory>
+            <directory>${project.build.outputDirectory}</directory>
+        </fileSet>
+    </fileSets>
+</assembly>
--- a/jte-gradle-plugin/pom.xml
+++ b/jte-gradle-plugin/pom.xml
@@ -125,6 +125,17 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>gg.jte.gradle</Automatic-Module-Name>
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
                 <version>3.3.0</version>
                 <configuration>
--- a/jte-gradle-plugin/pom.xml
+++ b/jte-gradle-plugin/pom.xml
@@ -113,21 +113,20 @@
     </repositories>
 
     <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-assembly-plugin</artifactId>
+                    <version>3.3.0</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
         <plugins>
-            <!-- Disable default JAR plugin -->
-            <plugin>
-                <artifactId>maven-jar-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>default-jar</id>
-                        <phase>none</phase>
-                    </execution>
-                </executions>
-            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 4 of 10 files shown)


pom.xml
```diff
@@ -51,6 +51,7 @@
         <module>jte-runtime-cp-test</module>
         <module>jte-test-report</module>
         <module>jte-deploy-nexus</module>
+        <module>jte-gradle-plugin</module>
```
casid, line 54: I think we should move the module up to where the maven module is.

I'm not sure if it will be deployed to nexus correctly, when it is after `jte-deploy-nexus`.

jte-gradle-plugin/src/main/java/gg/jte/gradle/JteGradle.java
```diff
@@ -0,0 +1,12 @@
+package gg.jte.gradle;
+
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+
+public class JteGradle implements Plugin<Project> {
+    @Override
+    public void apply(Project project) {
+        project.getTasks().register("generatorMojo", GeneratorMojoTask.class);
```
casid, line 9: Mojo is coming from maven. I'm not very proficient in gradle, so I'm not sure if there's something similar there. If not, we might just call it `generator`?

Hmm, or since it is named task, maybe just `generate` and `compile`?

casid (commented):
Wow, that looks pretty good, thank you for contributing a gradle plugin!

I started a review and left a few comments. Let me know what you think.

I noticed the build is failing for JDKs > 11. 
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.2.0:jar (attach-javadocs) on project jte-gradle-plugin: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1 - javadoc: error - The code being documented uses packages in the unnamed module, but the packages defined in http://jte.gg/jte/apidocs/ are in named modules.
```

I'm not sure, but it might be resolved by adding an automatic-module-name to your pom.xml:

```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-jar-plugin</artifactId>
    <configuration>
        <archive>
            <manifestEntries>
                <Automatic-Module-Name>gg.jte.gradle</Automatic-Module-Name>
            </manifestEntries>
        </archive>
    </configuration>
</plugin>
```

jte-gradle-plugin/src/main/java/gg/jte/gradle/JteGradle.java
iRebbok: It's a task name, from gradle we can access it via `tasks.name`, like `tasks.generatorMojo`, I think it's not a best idea to give a task such a general name. I've looked at the source code of the maven plugin, now I think it was, uh, bad. Don't you think `generateJte` & `compileJte` would be better? I guess the task name should follow the package name, but I'm not sure tbh.

pom.xml
iRebbok: Sure, I didn't know the details.

iRebbok:
> I noticed the build is failing for JDKs > 11.
> 
> ```
> Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.2.0:jar (attach-javadocs) on project jte-gradle-plugin: MavenReportException: Error while generating Javadoc: 
> Error:  Exit code: 1 - javadoc: error - The code being documented uses packages in the unnamed module, but the packages defined in http://jte.gg/jte/apidocs/ are in named modules.
> ```
> 
> I'm not sure, but it might be resolved by adding an automatic-module-name to your pom.xml:
> 
> ```
> <plugin>
>     <groupId>org.apache.maven.plugins</groupId>
>     <artifactId>maven-jar-plugin</artifactId>
>     <configuration>
>         <archive>
>             <manifestEntries>
>                 <Automatic-Module-Name>gg.jte.gradle</Automatic-Module-Name>
>             </manifestEntries>
>         </archive>
>     </configuration>
> </plugin>
> ```

I used `maven-assembly-plugin` there because gradle plugins should contain their dependencies inside (it just couldn't even define the tasks due to missing dependencies). I'm also running into an issue with the `MANIFEST-INF/maven/gg.jte.gradle` folder, it's just won't be generated, I don't know why. I'll include the manifest entry into the `maven-assembly-plugin` configuration instead, I've disabled `maven-jar-plugin` so it doesn't any do unnecessary work, and `maven-assembly-plugin` replaces it.

iRebbok:
Oh, I already do that with `maven-assembly-plugin`. I dunno what's causing failing the builds to fail.
https://github.com/casid/jte/blob/534c0608997d16d3fb9c88eda880d986aeae4f5e/jte-gradle-plugin/pom.xml#L131-L143

iRebbok:
All the tests passed. Fine. But remember, `jte-gradle-plugin` requires packed dependencies into its `.jar`, otherwise it won't work as a plugin with gradle.

casid:
Thanks! I'm merging this now and will try to add a test gradle project.

jte-gradle-plugin/src/main/java/gg/jte/gradle/JteGradle.java
casid: Agreed! I renamed the tasks as you suggested: https://github.com/casid/jte/commit/82ba9d93d3c59fd75b2a6702c1aec251df3ed1a4

casid:
@iRebbok I worked on the gradle plugin this morning. Phew, gradle is really tough to get into, at least for me.

I think it may be better to build the gradle plugin using gradle. Apparantly there is a own repository, just for gradle plugins. I'll try to register there this weekend, so that we can publish the plugin.

I added a new module `jte-runtime-test-gradle`, that builds with the gradle plugin and then executes some tests.

Let me know what you think!

iRebbok:
> I think it may be better to build the gradle plugin using gradle. Apparantly there is a own repository, just for gradle plugins. I'll try to register there this weekend, so that we can publish the plugin.

I'm sure it's a great idea. I've spent about 2 hours working with maven and trying to build the gradle plugin, because [that stackoverflow post](https://stackoverflow.com/questions/27555560/how-can-i-build-a-gradle-plugin-with-maven) is too old and only helped with a link to the gradle repo.

casid:
Yeah, whenever I touch gradle it takes me hours to get into it again :-D

I think I got it running now. I added two test projects for gradle (basically copied the existing maven test projects).

We now have:
- `jte-runtime-test-gradle`: Runs the gradle plugin task `precompileJte`
- `jte-runtime-cp-test-gradle`: Runs the gradle plugin task `generateJte`

The gradle plugin and both test projects are now executed via Github Action after the maven build is finished. So we test the gradle plugin on all supported operating systems and all supported JDKs. Finally got all of them green 👍 : https://github.com/casid/jte/runs/1894268122

Would you like to have a look at the example projects and if those are gradle-like? If you give you go, I'd then try publish the gradle plugin to the official repo and extend the documentation.

Just want to say - thanks so much for starting this! I think having a gradle plugin will be really nice for other devs out there :-)

iRebbok:
Eh? Where are they located at?

casid:
Here:

https://github.com/casid/jte/tree/master/jte-runtime-test-gradle
https://github.com/casid/jte/tree/master/jte-runtime-cp-test-gradle

I already merged your PR and did those on the master branch. :-)

iRebbok:
Yep, heh, they are gradle-like.

casid:
Thanks!

```
Publishing plugin gg.jte.gradle version 1.6.0
Thank you. Your new plugin gg.jte.gradle has been submitted for approval by Gradle engineers. The request should be processed within the next few days, at which point you will be contacted via email.
Publishing artifact build/libs/jte-gradle-plugin-1.6.0.jar
Publishing artifact build/libs/jte-gradle-plugin-1.6.0-sources.jar
Publishing artifact build/libs/jte-gradle-plugin-1.6.0-javadoc.jar
Publishing artifact build/publications/pluginMaven/pom-default.xml
Activating plugin gg.jte.gradle version 1.6.0
```

Let's see how it goes :-D

Jerbell:
Well done.  Personally I find Gradle so much more fun than Maven.  :)  One day you'll convert the whole project. ;)

PR #64, opened 2021-02-27
docs: Markdown fixes + customizing class directory

IvanPizhenko:
I've fixed some general markdown syntax (VS Code markdownlint complained too much) and added small sub-chapter about customizing class directory and its influence on the hot reload.


casid:
Thank you! Markdown looks a lot cleaner now 👍

PR #67, opened 2021-03-09
Remove dependency on com.sun.nio.file.SensitivityWatchEventModifier

rferreira:
* com.sun has been moved to a jdk.unsupported module post jdk9 and causes compilation problems (without exposing the module) on jdk15. Based on the discussion here: https://github.com/apache/zookeeper/pull/1269/files it appears that its usefulness is questionable.
* other changed files are simply import optimizations handled by intellij


rferreira:
nm it looks like that SensitivityWatchEventModifier was doing something on macOS.

casid:
When I remember correctly I needed this flag for reliable file watching on my Macbook. I use jte on a site with pre-rendered templates and use this watcher for changes on template files.

I wasn't aware that `SensitivityWatchEventModifier` causes trouble on modularized projects, but it makes a lot of sense that the `com.sun` isn't part of any default modules anymore.

I think 99% of jte users won't need this ability of the directory resolver.

Maybe we could move this to another module and clean the main jte module from this dependency?

rferreira:
Hey this is great, I was going to try to refactor it like you suggested but you beat me to it. Looks like test coverage dropped for IOUtils so I whipped out some tests to see if it makes code coverage happy. 

Cheers

rferreira:
I'm not sure what to do about the codecov one, I removed the single if statement from the IoUtils deleteDirectoryContent and it is still failing :(

jte/src/main/java/gg/jte/compiler/IoUtils.java
```diff
@@ -18,17 +21,14 @@ public static String toString(InputStream inputStream) throws IOException {
     }
 
     public static void deleteDirectoryContent(Path directory) {
-        if (!Files.exists(directory)) {
-            return;
-        }
-
-        try {
+        Objects.requireNonNull(directory);
+        try (Stream<Path> pathStream = Files.walk(directory)) {
```
casid, line 25: Good call for properly auto-closing the stream.

jte/src/main/java/gg/jte/compiler/IoUtils.java
```diff
@@ -18,17 +21,14 @@ public static String toString(InputStream inputStream) throws IOException {
     }
 
     public static void deleteDirectoryContent(Path directory) {
-        if (!Files.exists(directory)) {
```
casid, line 21: I think we should keep this check. If there's no directory to delete, we don't need to try stuff and raise an exception.

jte/src/main/java/gg/jte/compiler/IoUtils.java
```diff
@@ -18,17 +21,14 @@ public static String toString(InputStream inputStream) throws IOException {
...
+        Objects.requireNonNull(directory);
+        try (Stream<Path> pathStream = Files.walk(directory)) {
             //noinspection ResultOfMethodCallIgnored
-            Files.walk(directory)
-                    .filter(d -> d != directory)
+            pathStream.filter(d -> d != directory)
                     .sorted(Comparator.reverseOrder())
                     .map(Path::toFile)
                     .forEach(File::delete);
+        } catch (NoSuchFileException ignored) {
```
casid, line 31: `NoSuchFileException` looks a bit suspicious here. Maybe have a `deleteQuietly` method that is called for each file?

jte/src/test/java/gg/jte/compiler/IoUtilsTest.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+class IoUtilsTest {
+    @Test
+    void shouldIdentifyTemplateFiles() {
+        Assertions.assertTrue(IoUtils.isTemplateFile("foo.jte"));
```
casid, line 16: Can you adjust the assertions to use AssertJ? That's what is currently used for all jte tests and it's a bit confusing if assertions are mixed throughout the project.

jte/src/main/java/gg/jte/compiler/IoUtils.java
rferreira: Technically it won't since I'm catching the exception and ignoring it. I did this to try to lower the cyclomatic complexity but, as you can see, it didn't really work - happy to revert back.

jte/src/main/java/gg/jte/compiler/IoUtils.java
rferreira: I'll just revert this

jte/src/test/java/gg/jte/compiler/IoUtilsTest.java
rferreira: Yup yup

rferreira:
thanks again @casid and I just pushed a new commit with the tweaks you suggested.

casid:
@rferreira nice, that looks good!

PR #68, opened 2021-03-10
Gradle path fix

edward3h:
I came across a build issue when trying to run the gradle tests in a dev environment. It doesn't seem to happen in the clean environment of the github workflow.

The fix is a minor change to build.gradle files.
I've added the shell script I was using to run all the tests locally.

Side note: https://github.com/nektos/act is an alternative way to run tests locally, with its own limitations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,10 @@ Temporary Items
 /jte-runtime-test-kotlin/jte-classes/
 /jte-runtime-test-gradle/jte-classes/
 /jte-hotreload-test/jte-classes/
+
+### VS Code / VS Code Java extension (uses Eclipse code) ###
+.project
+.settings/
+.classpath
+bin/
+.factorypath
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -645,7 +645,7 @@ tasks.generateJte {
     contentType = ContentType.Html
 }
 
-sourceSets.main.java.srcDirs += tasks.generateJte.targetDirectory
+sourceSets.main.java.srcDir(tasks.generateJte.targetDirectory)
 
 tasks.compileJava {
     dependsOn(tasks.generateJte)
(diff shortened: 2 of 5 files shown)


casid:
Thanks @edward3h!

I must admit, that my Gradle knowledge isn't very deep.

What's the difference between `java.srcDir` and `java.srcDirs`? I like that the Kotlin DSL and Groovy build files are very similar after your changes. Would still be interested to know what's going on :-)

I think we should also update the samples in the documentation: https://github.com/edward3h/jte/blob/gradle-path-fix/DOCUMENTATION.md#using-the-application-class-loader-since-120

edward3h:
I am the opposite - I've worked on Gradle plugins but know little about Maven...

This is the javadoc for the type 'sourceSets.main.java' is referring to in build.gradle
https://docs.gradle.org/current/javadoc/org/gradle/api/file/SourceDirectorySet.html

The problem was use of "+=" operator, combined with confusing API definition here (what's the difference between getSrcDirs() and getSourceDirectories() ? It turns out quite a lot)

If you interpret the Groovy property references and operator overloading the line
`java.srcDirs += x`
becomes something like
```
Set<File> files = new HashSet<>(java.getSrcDirs())
files.add(x)
java.setSrcDirs(files)
```
Since the getSrcDirs() method doc says it doesn't filter and it forces a calculation of the contents, this is a problem. I guess that it is an older API that has been left for backwards compatibility. The newer Gradle style is to use wrapper types that defer calculation of the result until it is needed for actually running the task.

So anyway calling the 'srcDir(x)' method just tells the object to add the path, which is what we want. 'srcDirs(x)' would do exactly the same, it's just a version that can take multiple objects. I preferred the singular since there's only one in this case.

Also, I'll update the documentation page :)

edward3h:
I'm hoping to enhance the Gradle plugin so that the project including the plugin doesn't have to do the sourceSet manipulation, by the way. But not in this PR.

casid:
Oh wow! Thanks for the in-depth explanation! That actually makes a lot of sense and helped me grasp gradle a little bit better.

I'm merging this PR now.

Also - making the plugin usage simpler by getting rid of the sourceSet manipulation sounds great!

PR #81, opened 2021-05-06
Gradle plugin improvements

edward3h:
Here's the Gradle improvement I was thinking about. 

Reduce the amount of boilerplate when using the Gradle plugin by adding an extension with properties and default values.


casid:
Holy shit @edward3h, that looks fantastic!

```
jte {
  precompile()
}
```

Can't really get much nicer than this! I'll look through the changes this weekend 👍

edward3h:
The reason there are so many files and changes is because I copied the existing gradle test projects to make the same tests for the new setup. There's probably a way to refactor and reduce duplication, but I'm not planning to work on that now.

casid:
@edward3h yeah, I've seen it. I think duplication is perfectly fine here, maybe one day we can remove the old gradle style tests altogether.

casid:
Merged! 

Thanks for putting in so much effort in this, this is an incredible improvement of the gradle plugin.

That being said, I think it is time for a new release, with native image support and the gradle extension :-)

PR #109, opened 2021-07-18
Pull request for issue 104: putParametersOnSeparateLines option was implemented.

izogfif:


jte-runtime/src/main/java/gg/jte/runtime/StringUtils.java
```diff
@@ -70,4 +70,22 @@ public static void appendEscaped(StringBuilder result, String string) {
             }
         }
     }
+
```
casid, line 73: Can you please move these methods to a StringUtils class in the JSP converter project? It is not required by jte itself.

I know we can't use `String.repeat`, because of Java 8 compatibility.

jte-jsp-converter/src/test/java/gg/jte/convert/jsp/JspToJteConverterTest.java
```diff
@@ -244,12 +251,17 @@ void givenSetup(Consumer<Converter> setup) {
         setups.add(setup);
     }
 
-    void whenJspTagIsConverted(String jspTag, String jteTag) {
+    void whenJspTagIsConverted(String jspTag, String jteTag, boolean putParametersOnSeparateLines) {
```
casid, line 254: Do we need a parameter for this? I think its a sane default to format the parameters like you did. As @kelunik pointed out if someone requires different formatting, this should be done through a formatter. Conversion is a one time job, the need to format jte files will come after that many more times.

Right now you test only that this usecase works. All other conversion tests might work, but we're not sure and we don't know how they look like with this parameter on.

casid (changes requested):
Left some comments

izogfif:
@casid I've made changes as requested. `StringUtils.repeat` method was removed and inlined. `JspToJteConverterTest.whenJspTagIsConverted` now always enables `putParametersOnSeparateLines`. I've updated some tests, found the bug (when the tag has no parameters, but has body, there was an error in formatting) and fixed it.

jte-jsp-converter/testdata/simpleTagWithBody2/after/jte/tag/simple.jte
```diff
@@ -2,6 +2,9 @@
...
 @param String title
 
-@tag.withBody(headline = title, bodyContent = @`
+@tag.withBody(
+    headline = title,
+    bodyContent = @`
     <p>This is the body, for title ${title}</p>
-`)
\ No newline at end of file
+`
```
casid, line 9: The closing content backtick and the closing brace should be on the same line.

jte-jsp-converter/src/test/java/gg/jte/convert/jsp/JspToJteConverterTest.java
```diff
@@ -247,6 +254,7 @@ void givenSetup(Consumer<Converter> setup) {
     void whenJspTagIsConverted(String jspTag, String jteTag) {
         JspToJteConverter converter = new MyConverter();
         converter.convertTag(jspTag, jteTag, c -> {
+            c.setPutParametersOnSeparateLines(true);
```
casid, line 257: This should be the default setting for the converter. Otherwise we're now having tests for a configuration the default user does not experience, which is bad.

jte-jsp-converter/testdata/simpleTagWithBody3/after/jte/tag/simple.jte
```diff
@@ -2,6 +2,8 @@
 
 @param String title
 
-@tag.withBody(bodyContent = @`
+@tag.withBody(
+    bodyContent = @`
     <p>This is the body, for title ${title}</p>
-`)
\ No newline at end of file
+`
```
casid, line 8: Same here...

casid (changes requested):
Thanks for the adjustments @izogfif. I added some more comments.

jte-jsp-converter/testdata/simpleTagWithBody2/after/jte/tag/simple.jte
izogfif: @casid
I had an impression that closing brace should always be on a separate line, except when there are no parameters at all. Samples:
```
@tag.withoutParameters()
@tag.withOneParameter(
    param1 = "value1"
}
@tag.withContentParamNoLineBreaks(
    content1 = @`<div>Text</div>`
)
@tag.withContentParamNotEndingWithLineBreak(
    content1 = @`<div>
    <span>text</span>
</div>`
)
@tag.withContentParamEndingWithLineBreak(
    content1 = @`<div>
    <span>text</span>
</div>
`
)
@tag.withContentParamFollowedByIntParam(
    content1 = @`<div>
Text
</div>
`,
    name = "John"
)
@tag.withTwoContentParams(
    content1 = @`<span>Text 1</span>
`,
    content2 = @`<span>Text 2</span>`
)
@tag.withTwoContentParamsBothEndingWithLineBreak(
    content1 = @`<p>John</p>
`,
    content2 = @`<p>Smith</p>
`
)
```

jte-jsp-converter/src/test/java/gg/jte/convert/jsp/JspToJteConverterTest.java
izogfif: Should I remove the `setPutParametersOnSeparateLines` method in the first place? Or keep it so that the end user can turn it off (it's on by default)?

casid: I would remove the option. The converter may come with a sensible default setting how stuff is formatted (now, tag parameters on new lines), but it is not a formatter, so we should not put too much formatting logic in here. If different formatting is required, this effort is better put in the IntelliJ plugin.

jte-jsp-converter/testdata/simpleTagWithBody2/after/jte/tag/simple.jte
casid: Hmm, especially your last example looks a bit assymetric, doesn't it?

```
@tag.withTwoContentParamsBothEndingWithLineBreak(
    content1 = @`<p>John</p>
`,
    content2 = @`<p>Smith</p>
`
)
```

This would be more consistent:

```
@tag.withTwoContentParamsBothEndingWithLineBreak(
    content1 = @`<p>John</p>
`,
    content2 = @`<p>Smith</p>
`)
```

jte-jsp-converter/src/test/java/gg/jte/convert/jsp/JspToJteConverterTest.java
kelunik: Rather there should probably be a command line tool for formatting instead of relying on the plugin, so it can also be applied in CI?

jte-jsp-converter/testdata/simpleTagWithBody2/after/jte/tag/simple.jte
izogfif: OK, leaving it as it is.

izogfif:
@casid I've removed `putParametersOnSeparateLines` setting. Code was adjusted as if this setting is always enabled.

jte-jsp-converter/src/test/java/gg/jte/convert/jsp/JspToJteConverterTest.java
izogfif: > Rather there should probably be a command line tool for formatting

Good point, but it's outside of the scope of this issue. Option removed.

casid (approved):

casid:
Thanks @izogfif, merged.

Issue #125, opened 2021-12-18
Question: How to escape the template engine syntax?


renannprado:
For example, let's say I have an HTML page that has some inline javascript, like the below one.

```javascript
console.log(`testing ${myVar}`);
```
How can I escape the `jte` syntax so to say, in that piece of code, to make sure it won't be processed?

kelunik:
```jte
console.log(`testing ${"${myVar}"}`);
```

Usually using a string is the easiest way, I think.

renannprado:
@kelunik thanks. That would work. However, it would be really nice if there was a real escape feature.

renannprado:
@casid so if I understood correctly, there won't be a way to do the escape then? I mean, when we mix jte with some more complex things like react or vue, couldn't it be it would be more problematic?

casid:
@renannprado I don't think it is worth the extra complexity. Also, I don't think jte will be used to generate vue/react files, right?

For the Javascript example above, it would also be possible to use a unicode escape sequence.

renannprado:
> Also, I don't think jte will be used to generate vue/react files, right?

@casid  I'm not sure if that would be realistic tbh. Let's say it's not. 

Still I do believe it's realistic to mix _javascript_ in the same file (even if it's small bits) or could be that I'm old-fashioned and like to use jquery with the dollar sign. You can argue of course that these things belong to a different non-jte-file, that would be a workaround/solution as well. `jte` is great, really. Maybe my examples are not convincing enough, but a do believe that would a nice addition to the engine. 

> I don't think it is worth the extra complexity

I understand what you're saying here. Do you think we could come up with a design that doesn't add too much complexity?

Without knowing the codebase, and whether that would be too complex to implement, I would imagine something like the below (as opposed to the escape sequence I suggested in the beginning):

```js
// the below code is not processed by JTE
@skip
const myVar = "abc123";
console.log(`test ${myVar}`);
@endskip
```
or even based on "annotations"/comments (maybe this one would be more difficult):

```js
// the below code is not processed by JTE
// @jteskip
const myVar = "abc123";
console.log(`test ${myVar}`);
// @endjteskip
```

If you're still not convinced, no problem :smile: , there are workarounds anyway.

casid:
For the escape sequence I meant something like this:

```javascript
const myVar = "abc123";
console.log(`test \x24{myVar}`);
```

[...]

PR #126, opened 2022-01-15
Template calls

casid:
This is a generalization so that every template can be called from any template.

Tags and Layouts are templates like any other and should be refactored to a more generic behavior.

This feature allows to structure code freely.

For example, page template located at `module1/page1.jte` could extract templates that are only used by this page. Those templates could be now located right next to that page, like this:

`module1/page1-component1.jte`
`module1/page1-component2.jte`

The calls from `module1/page1.jte` would then look like this:

```html
@template.module1.page1-component1()
@template.module1.page1-component2()
```

This is a breaking change and requires a bump to version 2.0.0.

To mitigate the inconvenience for users, the template compiler should be adjusted:
- Fail hard when `@tag.` or `@layout.` is detected when compiling a template
- Generate an error message containing Java code, that the user can use to automatically migrate the currently compiled project


jte/src/test/java/gg/jte/TemplateEngineTest.java
```diff
@@ -543,6 +543,132 @@ void tagWithVarArgs4() {
...
+    @Test
+    void escapeAt_templateCall() {
+        givenTemplate("mail@@example.com(write me)"); // Escaped with double @
+        thenOutputIs("mail@example.com(write me)");
+    }
+
+    @Test
+    void escapeAt_3x() {
+        givenTemplate("@@@");
+        thenOutputIs("@@");
```
kelunik, line 603: This should throw a syntax error IMO, so all `@` characters are reserved and have to be escaped or might get used in keywords in later releases. That way, there won't be BC breaks if new keywords starting with `@` are introduced.

jte/src/test/java/gg/jte/TemplateEngineTest.java
```diff
@@ -543,6 +543,132 @@ void tagWithVarArgs4() {
...
+        givenTemplate("@parameter(model.hello, model.x), That was an arbitrary template call!");
+        thenOutputIs("One: Hello, two: 42, That was an arbitrary template call!");
+    }
+
+    @Test
+    void templateCall_noConflictWithKeywords_import() {
+        dummyCodeResolver.givenCode("importer.jte", "@param java.lang.String firstParam\n" +
+                "@param int secondParam\n" +
+                "One: ${firstParam}, two: ${secondParam}");
+        givenTemplate("@importer(model.hello, model.x), That was an arbitrary template call!");
```
kelunik, line 578: I think we shouldn't allow arbitrary words right after `@`, as that limits extending the library with new keywords without BC breaks. Maybe something like `@include.importer`?

casid: Good point. The reason for arbitrary words is mainly to keep `@tag` and `@layout` working without special handling.

casid: As we discussed today, I've changed the keyword to `@template`. Thanks for the suggestions, I really like where this is going!

jte/src/test/java/gg/jte/TemplateEngineTest.java
casid: With the new keyword we don't absolutely need escaping for this PR, so I removed it for now. You're right, and the current implementation is a little bit too rushed to be any good :-D.

Will continue this in https://github.com/casid/jte/issues/125

PR #153, opened 2022-05-12
Native build update

edward3h:
Replaced the palantir graal gradle plugin with the GraalVM buildtools plugin. This has built in support for running tests in a native build, so eliminates my own previous workaround.

Added a separate Github Workflow for running tests in GraalVM so it can use the graal action instead of normal JDK. The Windows builds are currently failing with a cause that looks similar to https://github.com/graalvm/native-build-tools/issues/234 so I removed Windows as a target for now.


casid:
That's really nice, thank you!

To speed up the workflow duration and to prevent running tests twice, I think we should run maven on the native action with `skipTests=true`. I'm not sure about the gradle tests but maybe we can get rid of some redundancy there as well?

edward3h:
Yes, I removed the gradle tests that aren't doing native image so the time should be reduced a bit now.

casid:
Thank you!

PR #155, opened 2022-05-18
fix for a problem I encountered using binary static content

edward3h:
with generate in gradle build - missing resources directory.

Let's see if the workflow tests complete ok ;)


edward3h:
So the workflow tests didn't :( It worked 'on my computer'. 
Will come back to this later in the week.

edward3h:
Doh! The issue the workflow is failing on now is literally something I didn't implement in the native build support. It needs some additional configuration to tell it to load the `.bin` files.

edward3h:
@casid a question about the binary support - why are the binary parts written to a separate file rather than e.g. byte arrays in the generated classes?

casid:
> @casid a question about the binary support - why are the binary parts written to a separate file rather than e.g. byte arrays in the generated classes?

Byte array definitions are quite verbose. This is done to keep the class files small.

casid:
Thank's for the fix!

PR #165, opened 2022-06-27
Add the ability to override the jte template suffix (so .kte templates are supported in spring boot)

barthr:
It was only possible to compile jte templates in the spring boot starter, this PR makes sure that you can override that default to also allow for kte templates


casid:
Thanks!

PR #186, opened 2022-11-19
Implement ResourceCodeResolver.getLastModified

andreblanke:
Hot reloading is currently unsupport for `ResourceCodeResolver` due to `getLastModified` always returning `0`. By implementing the method on a best-effort basis, hot reloading can be supported when the resource URL points to a regular file outside a JAR file which is usually the case during development time. A project which uses jte can then be rebuilt which will cause the changes to be reflected.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -439,7 +439,7 @@ Sometimes it is required to output text as is. The `@raw` keyword can be used to
 
 ### For a regular website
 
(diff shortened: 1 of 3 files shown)


edward3h:
I ran into this issue before. Looks like a reasonable fix. 👍🏻

casid:
@andreblanke thanks for the contribution! Right now, the `ResourceCodeResolverTest` fails, cause it isn't expecting this to work yet. Could you adjust the test and maybe add another test for the unhappy path, too?

andreblanke:
Ah, sorry for that. I tested it locally but didn't actually check the tests. I will fix it.

PR #188, opened 2022-11-24
Typo: directoy -> directory

rferreira:
Just a very minor typo


casid:
Thanks for the fix! 👍

PR #195, opened 2023-02-01
add dedicated starter for spring boot 3

### LINKED ISSUE #193: Jakarta Support for SpringBootStarter
Hello, I would love to use JTE in my new project with Spring Boot 3, but JTE (jte-spring-boot-starter) 2.2.4 only supports javax as I can see. 

When will jakarta namespace be supported?

atomfrede:
I need to test both starters with a sample project, so this is in draft mode, but I don't see why it should not work :)

closes #193


atomfrede:
Ah and of course boot 3 requires java 17 as baseline.

atomfrede:
@casid What would you say is the best way to handle the jdk8/11 vs jdk17 baseline? For e.g. mongock (mongo db migration) starter we created a dedicated artifact (with repository etc.) to keep the core compatible with jdk11 while the starter has jdk17 as baseline. Or should we try keep in one artifact and only compile the starter-3 with 17?

casid:
I think the best would be to run maven with `-pl !jte-spring-boot-starter-3` in case `matrix.java_version` is one of `1.8` or `11`.

https://github.com/casid/jte/blob/main/.github/workflows/maven.yml#L26

atomfrede:
Indeed sounds very easy. How is the release process eventually?

casid:
Thanks, that looks good!

As for the release, I will need to use JDK 17 from now on (used 11 before), but that should not be a problem. 

Let me know when you want this to be merged :-)

atomfrede:
Need to check a little further. Something seems to be broken

atomfrede:
@casid Ready to go.

casid:
Sorry for the late reply & merge! Thank you @atomfrede!

Issue #204, opened 2023-03-21
Gradle Plugin: Support configuration cache


nbrugger-tgm:
Starting with Gradle 8.1 the configuration-cache is not unstable anymore 🥳 

At the moment the jte gradle plugin completely halts the gradle process when the config cache is enabled

```
* What went wrong:
Execution failed for task ':core:generateJte'.
> Cannot query the value of this property because it has no value available.
```

```
org.gradle.unsafe.configuration-cache=true
```

edward3h:
8.1 is a release candidate at the moment? Thanks for the heads up.

I'm kinda busy this week but can try and find time to look into it next week.

nbrugger-tgm:
Yes it is RC right now so its not too pressing, thank you for taking a look into it 👍🏻

edward3h:
@casid Do you have any preference for which Gradle versions to support? I would probably try and keep backwards compatibility with 7.x, though I know there were some discussions about Kotlin version support that were related.

casid:
@edward3h just came back from vacation, so sorry for the late reply. You're right, for Kotlin/Gradle the minimum Gradle version is now 8.x. Other combinations currently still work with Gradle < 8.x, so it's nice that you added support :-)

Also, thanks for the PR, I'd have no idea what to do here!

nbrugger-tgm:
Is the feature already released? If yes this ticket could be closed with a version note.

Thank you for your effort

casid:
@nbrugger-tgm I just published [version 2.3.1](https://github.com/casid/jte/releases/tag/2.3.1), that includes this feature.

Thank you @edward3h, for making this happen :-)

PR #207, opened 2023-04-02
Support Gradle configuration cache.

edward3h:
Fix issues related to Gradle configuration cache. #204 

Add gradle-test-wrapper to support running Gradle tests against multiple Gradle versions.


casid:
Thanks @edward3h!

This looks good and I would like to merge it, but the build got a lot slower.

_Build with maven_: 3-5 min (unchanged and unrelated. I'd still like to have this within 1 minute, but most of the time is eaten by the slow Kotlin compiler)
_Run all the Gradle Plugin tests_: 16-25 min (this seems quite slow, essentially the tests don't do much at all, I'd expect each to be done within a few seconds)

I feel we should try to speed up the Gradle tests before merging. Do you have an idea what could cause the slowdown?

edward3h:
I made it run each test multiple times with different Gradle versions. I'll add a property to control that. I'm also not sure if running with TestKit added overhead. I'll take a look at timing.

edward3h:
As a guess it was downloading and installing each gradle version in the CI build. I'm making it default to just running with the current version.

casid:
Oh wow, that's much faster now! Thanks for looking into it, and also for contributing the Gradle configuration cache support.

PR #208, opened 2023-04-02
Add the '@Generated' annotation to generated code

edward3h:
Not essential, but it is common to annotate generated code. It also makes it easier to make an annotation processor use the classes.

[javadoc for @Generated](https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/annotation/processing/Generated.html)


edward3h:
Now I realize '@Generated' was added in JDK 9, so isn't available since JDK 8 is the target for this project.

kelunik:
Even premier support for JDK 8 ended more than a year ago: https://www.oracle.com/java/technologies/java-se-support-roadmap.html

I think it's time to drop JDK 8 support if there are features we'd like to use.

casid:
Agreed. Should we go the Spring way and release jte 3.0.0, with JDK 17 as baseline?

There could even be a few small performance improvements (when I added Java 8 support, I think the `DirectoryCodeResolver`  got a bit slower because it couldn't use the `Files` API (7fe0864f48de426c30b37fdeb858840327547c79)

edward3h:
I can't speak for anyone else, but my usage of jte is all on JDK 17+ now.

I know Micronaut are planning to release a 4.x with JDK 17 as baseline.

agentgt:
FWIW I think Generated provides very little value and if you add the date attribute you break reproducible builds.

I tried looking for tools that used it and did not find many  when I was making my project.

What I instead provide is an option for the user to put any annotation they want on generated classes. See https://jstach.io/doc/jstachio/current/apidocs/io.jstach.jstache/io/jstach/jstache/JStacheInterfaces.html#templateAnnotations()

BTW Doesn’t JTE just generate bytecode directly or is this some of the newer type safe model work?

edward3h:
`@Generated` is documentation that can be used by tools. I accept that in practice few tools actually use it. The date attribute could be dropped.

I'd like to add capability for a custom annotation. I think that makes sense on the jte-models module I just added, rather than the core render classes.

JTE generates source code, not bytecode. In the case where dynamic template reloading is enabled, it generates source then calls the Java compiler to compile it.

edward3h:
I see. I will drop this one.

agentgt:
@edward3h don't drop the change! I didn't mean to sound like a debbie downer. I think there is merit for adding some annotations just need to flesh it out more or as an option at least document that folks will need to special things for modules.

casid:
Reopened as I‘m not sure if this was closed for the right reason.

> Well in that case it is even worse because users would in theory now then need to put java.compiler.

@agentg is this because javax.annotation.processing.Generated is located in java.compiler module?

agentgt:
> @agentgt is this because javax.annotation.processing.Generated is located in java.compiler module?

Yes. 

If one is using modules and has a module-info.java then their effective visible symbols (aka when you press ctrl+space) will now include all of the code in `java.compiler`. Furthermore I believe annoyingly if one ships something packaged up using jlink the compiler module would have to be included but that might be the case regardless if JTE runtime needs it.

There is value in what @edward3h is proposing albeit I'm not sure it is the `@Generated` annotation and I think a general purpose put this annotation on the class would be useful.

For example depending on how JTE does its null checking one might want to have the `@NullByDefault` (or whatever analog for nullable annotation library) or perhaps some other annotation to ignore Javadoc generation. Those are two tools I know will look at derived sources but I don't think automatically use `@Generated`.

edward3h:
I've added https://github.com/casid/jte/pull/235 instead, which is about adding annotations to the jte-models facade classes.

PR #224, opened 2023-05-16
Generator extension API + jte-models extension

edward3h:
This is what I've done for generating a 'model' facade. I need to finish some more documentation, but thought it would be good to share.

* added an 'extension API' that can be called from TemplateCompiler
* jte-models module is an extension that generates a facade interface + static and dynamic implementations.
* maven and gradle plugin support
* test cases


casid:
@edward3h that‘s awesome progress! I‘ll have a detailled look this weekend and proper feedback 😊

jte-extension-api/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>gg.jte</Automatic-Module-Name>
```
casid, line 31: The module name should be `gg.jte.extension.api` in order to not conflict with the `gg.jte` module.

jte-maven-plugin/src/main/java/gg/jte/maven/GeneratorMojo.java
```diff
@@ -77,7 +83,13 @@ public void execute() {
             templateEngine.setGenerateNativeImageResources(generateNativeImageResources);
         }
         templateEngine.setProjectNamespace(project.getGroupId() + "/" + project.getArtifactId());
-
+        getLog().info("extensions=" + extensions);
```
casid, line 86: Should we move this into the if, so that it only logs if there are extensions configured?

jte-models/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>gg.jte</Automatic-Module-Name>
```
casid, line 63: The module name should be `gg.jte.models` in order to not conflict with the `gg.jte` module.

jte-native-resources/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>gg.jte</Automatic-Module-Name>
```
casid, line 31: The module name should be `gg.jte.native.resources` in order to not conflict with the `gg.jte` module.

jte/src/main/java/gg/jte/compiler/TemplateCompiler.java
```diff
@@ -298,7 +264,8 @@ public ClassInfo generateTemplateCall(String name, LinkedHashSet<ClassDefinition
         CodeGenerator codeGenerator = createCodeGenerator(classInfo, classDefinitions, templateDependencies);
         new TemplateParser(code, TemplateType.Template, codeGenerator, config, codeResolver).parse();
 
-        classDefinition.setCode(codeGenerator.getCode(), codeGenerator.getBinaryTextParts());
+        classDefinition.setCode(codeGenerator.getCode(), codeGenerator.getBinaryTextParts(), codeGenerator.getParamInfo(), codeGenerator.getImports());
+        classDefinition.setParams(codeGenerator.getParamInfo());
```
casid, line 268: I think the setParams() call is redundant here. Maybe we can remove that method?

jte/pom.xml
```diff
@@ -24,6 +24,16 @@
             <artifactId>jte-runtime</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>gg.jte</groupId>
+            <artifactId>jte-extension-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
```
casid, line 32: Should we remove the native resources extension here? This way users who need it can add the dependency and others don't have it by default?

jte-models/src/main/java/gg/jte/models/runtime/JteModel.java
```diff
@@ -0,0 +1,30 @@
...
+        Writer writer = new OutputStreamWriter(outputStream, charset);
+        render(writer);
+    }
+
+    default void render(OutputStream outputStream) {
+        render(outputStream, StandardCharsets.UTF_8);
+    }
+
+    default String render() {
+        Writer writer = new StringWriter();
```
casid, line 23: `StringWriter` uses a synchronized `StringBuffer` under the hood and is rather slow.

I'm wondering if it wouldn't be better to have only one `render` method that accepts a `TemplateOutput` argument? This way users have full control over rendering if they need to. For instance, if they use precompiled UTF-8 templates, or have a special need for output escaping.

We could then still offer a convenience `render` method that provides a String by using `StringOutput`.

casid (commented):
Hey @edward3h I finally had the time to do a proper review. And wow, that looks great! I really like that it's a general purpose extension API, as the native image resources are now possible with an extension instead of hacking into the project. Also, I find it really funny that you're using jte to generate the models :-D.

I left a few notes in the code, mostly minor stuff. I have left a note on the `JteModel` interface, I have a gut feeling that the render() methods there need to be simplified a bit. I left a suggestion, let me know what you think about it!

Overall I feel very confident that we can merge this soon.

jte-extension-api-mocks/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>gg.jte</Automatic-Module-Name>
```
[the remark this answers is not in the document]
edward3h: Ok. This is me not knowing Maven well and copy-pasting. WIll fix.

jte-maven-plugin/src/main/java/gg/jte/maven/GeneratorMojo.java
edward3h: Yes, will do.

jte/src/main/java/gg/jte/compiler/TemplateCompiler.java
edward3h: Yes, missed this in changing the above method.

jte/pom.xml
edward3h: The intention was to keep backward compatibility for users who have configured native resources with the existing boolean property. Maybe there is another way to do that. Do you have any ideas?

I can figure out how to do it in Gradle :)

jte-models/src/main/java/gg/jte/models/runtime/JteModel.java
edward3h: I did not realise it would use a StringBuffer! I will definitely change that.

I think using TemplateOutput + convenience method will still meet my goals - I will try it.

edward3h:
Yes I was playing with String.format to generate model classes, then thought it would be easier with some sort of template library, then "Oh, I have one right here". It is nice that it worked out.

jte-native-resources/pom.xml
edward3h: I had to change this because of 'native' being a keyword.

jte/src/main/java/gg/jte/compiler/TemplateCompiler.java
edward3h: Done

jte-models/src/main/java/gg/jte/models/runtime/JteModel.java
edward3h: Done

jte/pom.xml
casid: Maybe we could throw an Exception when the native resources boolean property is set to true? And put instructions into the exception message how to migrate to the new extension?

jte-models/src/main/java/gg/jte/models/runtime/JteModel.java
casid: That looks really nice!

jte/pom.xml
edward3h: I guess if we put this in 3.0 we can just drop the existing setting and be less concerned about compatibility.

casid: Yes, that makes sense 👍

edward3h:
I pulled updates from main and got tests passing again.

casid:
That looks really good! I don't see anything that should prevent us from merging :-)

@edward3h what do you think?

edward3h:
Looks good to me :)

casid:
🎉 Thank you for this amazing extension API contribution @edward3h!

Issue #225, opened 2023-05-17
UTF-8 output performance


agentgt:
Hi @casid I have been benchmarking my templating engine to be used in TechEmpowers benchmark (Jooby) which requires UTF-8 output with non latin-1 characters.

https://github.com/agentgt/template-benchmark

https://github.com/agentgt/template-benchmark/raw/utf8/results-utf8.png

The three templating engines that are able to do pre-encoding are JStachio (mine), JTE, and Rocker.
The also happen to be the fastest templating engines in the Java ecosystem for String output.

I cannot figure why JTE performs that much worse than Rocker and JStachio. I made sure to `setBinaryStaticContent` both in Maven and in Java (I know that is working because if I turn that off it gets far worse). 

Perhaps the custom Utf8Output is the culprit? Did you benchmark it? Both JStachio and Rocker don't bother doing all the chunking logic you have.

My other guess it is the escaping of which I turned on (all three have escaping turned on for this benchmark as TechEmpower requires it).

https://github.com/agentgt/template-benchmark/blob/utf8/src/main/java/com/mitchellbosecke/benchmark/JteUtf8.java

casid:
Thanks for reporting @agentgt!

> I have been benchmarking my templating engine to be used in TechEmpowers benchmark (Jooby) which requires UTF-8 output with non latin-1 characters.

String can represent non latin-1 characters. The motivation is rather to save time from encoding static template parts to UTF-8 only once, right? I found this part of the README a bit confusing.

I'm not at my dev pc right now, but your jte setup looks correct. I'll have a closer look this weekend.

`Utf8Output` aims to avoid holding a big byte array containing the entire response data in memory. Instead it references the big static chunks and creates little dynamic ones on the fly. Those chunks are then passed one by one to the output stream of the web server. In your benchmark the result is a big byte array, so I imagine in this case `Utf8Output` introduces some overhead over just writing everything into a byte array.

agentgt:
> String can represent non latin-1 characters. The motivation is rather to save time from encoding static template parts to UTF-8 only once, right? I found this part of the README a bit confusing.

Yeah the readme sucks ass as I wrote it very quick but the ultimate goal is to output UTF-8 which is not a Java String its a byte representation. Yes Java Strings can represent UTF-8 but most platforms allow you to write direct bytes. So yes this is about saving time.

String.getBytes() is insanely fast provided every single character is latin1 but in the real world that is not the case even here in the US as people use emojis.

[...]

PR #227, opened 2023-05-21
Replace owasp java encoder with faster implementation

casid:
Benchmarks as in #225 have shown that OWASP Java Encoder is slower than expected for output escaping. In order to use the faster methods of OWASP encoder, all template outputs had to provide an instance of the rather ancient and clumsy `Writer` interface, that really should not be part of the public API of `TemplateOutput`.

This PR replaces the four methods we currently use from the OWASP Java Encoder project by optimized methods that work directly on a `TemplateOutput`:
- `org.owasp.encoder.Encode.forHtmlContent` -> `gg.jte.html.escape.Escape.htmlContent`
- `org.owasp.encoder.Encode.forHtmlAttribute` -> `gg.jte.html.escape.Escape.htmlAttribute`
- `org.owasp.encoder.Encode.forJavaScriptBlock` -> `gg.jte.html.escape.Escape.javaScriptBlock`
- `org.owasp.encoder.Encode.forJavaScriptAttribute` -> `gg.jte.html.escape.Escape.javaScriptAttribute`

All methods should behave exactly like their OWASP counterparts.

After this patch is merged, jte will be free from external dependencies.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -356,7 +356,7 @@ In `Html` mode, user content `${}` is automatically escaped, depending what part
 
 ### HTML tag bodies
 
-User output in HTML tag bodies is escaped with `org.owasp.encoder.Encode#forHtmlContent` (`org.owasp.encoder.Encode#forHtml` before jte 1.5.0).
+User output in HTML tag bodies is escaped with `gg.jte.html.escape.Escape.htmlContent`.
 
 ```htm
 <div>${userName}</div>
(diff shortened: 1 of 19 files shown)


kelunik:
This likely needs to be 3.0 then, because of the breaking change to the interface?

casid:
> This likely needs to be 3.0 then, because of the breaking change to the interface?

I think you are right, there are quite a few things that accumulated for a new major version. I've created a ticket to collect all changes in one place: #230

PR #231, opened 2023-05-23
229 set java 17 as new baseline

casid:


kelunik (approved):

casid:
@edward3h I had to change the extension package names a little bit, to fix the split package conflict between the two modules. Are you fine with the result?

https://github.com/casid/jte/pull/231/commits/b437f507889782b7f56f9641857263dd87b08399

edward3h:
Yes that make sense.

PR #235, opened 2023-05-31
Models custom annotation

edward3h:
Adds some optional properties to the jte-models extension, to enable adding an annotation to the generated interface or implementation classes. Added to test cases and README.


edward3h:
Looks like I will want to be able to specify distinct values for static vs dynamic template classes

edward3h:
When the extension properties are changed in build.gradle, gradle does not detect that inputs have changed. i.e. generated files were not rebuilt.

Issue #243, opened 2023-07-04
[3.0.0] Suggestions for jte-models


chkl:
Congratulations for the new 3.0.0 release!

I'm currently giving it a test run and especially the jte-models extensions looks like a great addition.

I have two small suggestions (which I'd be happy to open PRs for):

1. ensure that the generated method names are valid Java. (Currently the file name is taken in verbatim, turning e.g. `user-dashboard.jte` to `public JteModel user-dashboard(...)` ). 
2. add some kind of configuration option to only add methods for _some_ of the jte files. Not every little component needs its own method - imho.

edward3h:
Yes, those are good ideas. I'd be happy for you to open PRs :)

casid:
Wow, that was quick. Thank you for the heads up Christian - and thank you for the PRs!

casid:
Pull requests are merged. Thanks everyone! Closing this issue now.

chkl:
Thanks y'all!

PR #244, opened 2023-07-04
better method names (#243 Part 1)

chkl:
The first part of #243


casid:
Thanks Christian for the quick PR!

My only nitpick would be, could we add tests for the method names?

Otherwise, it looks good to me! @edward3h, you know the model extension a lot better, I'd say you should have the final say before merge :-)

kelunik (approved):

edward3h:
I agree it could do with some tests. Otherwise it looks good to me.

casid:
Merged, thanks for the PR @chkl!

PR #245, opened 2023-07-04
add include and exclude patterns to jte-model (#243 Part 2)

chkl:
As suggested in #243

### DIFF
--- a/jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
+++ b/jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
@@ -8,8 +8,10 @@ import gg.jte.extension.api.TemplateDescription;
 
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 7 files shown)


casid:
Looks good to me! @edward3h, you know the model extension a lot better, I'd say you should have the final say before merge :-)

As with the other PR, maybe we could add a test for this, possible integration test? There are `test/jte-runtime-cp-test-models`(for Maven) and `test/jte-runtime-cp-test-models-gradle` (for Gradle), which could configured with such patterns?

jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
```diff
@@ -31,6 +33,15 @@ public JteExtension init(Map<String, String> value) {
     @Override
     public Collection<Path> generate(JteConfig config, Set<TemplateDescription> templateDescriptions) {
         TemplateEngine engine = TemplateEngine.createPrecompiled(ContentType.Plain);
+
+        Pattern includePattern = modelConfig.includePattern();
+        Pattern excludePattern = modelConfig.excludePattern();
+
+        var templateDescriptionsFiltered = templateDescriptions.stream() //
+                .filter(x -> includePattern == null || includePattern.matcher(x.fullyQualifiedClassName()).matches()) //
+                .filter(x -> excludePattern == null || excludePattern.matcher(x.fullyQualifiedClassName()).matches()) //
```
kelunik, line 42:
```suggestion
                .filter(x -> excludePattern == null || ! excludePattern.matcher(x.fullyQualifiedClassName()).matches()) //
```

kelunik:
Also, this PR includes both changes now. We should either remove them or close #244.

chkl:
> Also, this PR includes both changes now. We should either remove them or close #244.

Thanks @kelunik for reviewing it and good catch. I added the missing negation and dropped the commits that are already included in #244 .


I'm going to add tests for this and #244 later.

PR #246, opened 2023-07-08
Add CompositeCodeResolver added to allow template reading in several places

DevAlves1993:
I recently had a problem with JTE when using it with the Spring View Component library in a personal project.
I needed a CodeResolver that could read templates in different places.

I therefore propose this pull request so that jte can take care of this need.
For details on the difficulty I had see the following [issue](https://github.com/tschuehly/spring-view-component/issues/8)

### DIFF
--- a/jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
+++ b/jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
@@ -1,0 +1,68 @@
+package gg.jte.resolve;
+
+
+import gg.jte.CodeResolver;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Resolves template code from multiple other resolvers
+ */
+public class CompositeCodeResolver implements CodeResolver {
+
+    private final List<CodeResolver> codeResolvers;
+
+    public CompositeCodeResolver(List<CodeResolver> codeResolvers) {
+        this.codeResolvers = codeResolvers;
+    }
+
+    @Override
+    public String resolve(String name) {
+        var content = "";
+        for (CodeResolver codeResolver : this.codeResolvers) {
+            try {
+                String resolve = codeResolver.resolve(name);
+                if(Objects.nonNull(resolve)){
+                    content = resolve;
+                }
+            } catch (Exception ex) {
+                // ignore
+            }
+        }
+        if(!content.isEmpty())
+            return content;
+        throw new UncheckedIOException(new IOException("Could not find template " + name));
+    }
+
+    @Override
+    public long getLastModified(String name) {
+        long lastModified = 0;
+        for (CodeResolver codeResolver : this.codeResolvers) {
+            try {
+                lastModified = codeResolver.getLastModified(name);
+                if(lastModified > 0)
+                    return lastModified;
+            } catch (Exception ex) {
+                lastModified = 0;
+            }
+        }
+        return lastModified;
+    }
+
+    @Override
+    public List<String> resolveAllTemplateNames() {
+        List<String> allTemplateNames = new ArrayList<>();
+        for (CodeResolver codeResolver : this.codeResolvers) {
+            try {
+                allTemplateNames.addAll(codeResolver.resolveAllTemplateNames());
+            } catch (Exception ex) {
+                // ignore
+            }
+        }
+        return allTemplateNames;
+    }
+}
--- a/jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
+++ b/jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
@@ -3,11 +3,8 @@ package gg.jte.resolve;
 
 import gg.jte.CodeResolver;
 
-import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 
 /**
  * Resolves template code from multiple other resolvers
@@ -20,22 +17,24 @@ public class CompositeCodeResolver implements CodeResolver {
         this.codeResolvers = codeResolvers;
     }
 
+    public CompositeCodeResolver(CodeResolver... codeResolvers) {
+        this(List.of(codeResolvers));
+    }
+
     @Override
     public String resolve(String name) {
-        var content = "";
+        String content = null;
         for (CodeResolver codeResolver : this.codeResolvers) {
             try {
                 String resolve = codeResolver.resolve(name);
-                if(Objects.nonNull(resolve)){
+                if(resolve != null){
                     content = resolve;
                 }
             } catch (Exception ex) {
                 // ignore
             }
         }
-        if(!content.isEmpty())
-            return content;
-        throw new UncheckedIOException(new IOException("Could not find template " + name));
+        return content;
     }
 
     @Override
--- a/jte/src/test/java/gg/jte/resolve/CompositeCodeResolverTest.java
+++ b/jte/src/test/java/gg/jte/resolve/CompositeCodeResolverTest.java
@@ -1,0 +1,59 @@
+package gg.jte.resolve;
+
+import org.junit.jupiter.api.Assertions;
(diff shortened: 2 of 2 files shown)


jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Objects;
+
+/**
+ * Resolves template code from multiple other resolvers
+ */
+public class CompositeCodeResolver implements CodeResolver {
+
+    private final List<CodeResolver> codeResolvers;
+
+    public CompositeCodeResolver(List<CodeResolver> codeResolvers) {
```
edward3h, line 19: For ease of use I would add an overloaded constructor that takes a varargs parameter. i.e.

```java
public CompositeCodeResolver(CodeResolver... codeResolvers) {
    this(List.of(codeResolvers));
}
```

jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
```diff
@@ -0,0 +1,68 @@
...
+        this.codeResolvers = codeResolvers;
+    }
+
+    @Override
+    public String resolve(String name) {
+        var content = "";
+        for (CodeResolver codeResolver : this.codeResolvers) {
+            try {
+                String resolve = codeResolver.resolve(name);
+                if(Objects.nonNull(resolve)){
```
edward3h, line 29: `resolve != null` would be fine. `Objects.nonNull()` exists mainly to support Stream filters.

jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
```diff
@@ -0,0 +1,68 @@
...
+                    content = resolve;
+                }
+            } catch (Exception ex) {
```
edward3h, line 32: `CodeResolver` does not declare any checked exceptions. I don't think it is appropriate to be swallowing downstream exceptions here. Are you encountering exceptions in your usage?

jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
```diff
@@ -0,0 +1,68 @@
...
+                // ignore
+            }
+        }
+        if(!content.isEmpty())
```
edward3h, line 36: I'll defer to @casid on his preferred code style, but everything I work on requires braces for `if` and similar, so this looks weird to me.

edward3h:
I made some code comments, but in general I agree this is a good idea.

jte/src/main/java/gg/jte/resolve/CompositeCodeResolver.java
DevAlves1993: The following implementations of DirectoryCodeResolver and ResourceCodeResolver each throw exceptions.

DirectoryCodeResolver 

```java
    @Override
    public String resolve(String name) {
        try {
            Path file = root.resolve(name);
            return Files.readString(file, StandardCharsets.UTF_8);
        } catch (NoSuchFileException e) {
            return null;
        } catch (IOException e) {
            throw new UncheckedIOException(e);
        }
    }
```

ResourceCodeResolver

```java
 @Override
    public String resolve(String name) {
        try (InputStream is = getClassLoader().getResourceAsStream(root + name)) {
            if (is == null) {
                return null;
            }
            return IoUtils.toString(is);
        } catch (IOException e) {
            throw new UncheckedIOException(e);
        }
    }
```

But I don't have to throw an exception in the loop's catch clause. I propose this new implementation below

```java
    @Override
    public String resolve(String name) {
        var content = "";
        for (CodeResolver codeResolver : this.codeResolvers) {
            try {
                String resolve = codeResolver.resolve(name);
                if(resolve != null){
                    content = resolve;
                }
            } catch (Exception e) {
                // ignore
            }
        }
        if(!content.isEmpty())
            return content;
        return null;
    }
```

casid: As @edward3h suggested, it is not a good idea to swallow exceptions like this. Both code resolvers return `null` if the template does not exist. In case I'm loading a template that exists, but there's an IO error during loading, users want to know this happened.

casid:
Welcome @DevAlves1993 and thanks for the PR. So far, jte expects templates in one root directory.

Changing this to have multiple sources of templates will make a few things more complicated:
- the IntelliJ plugin does not know about this yet and would need to be adjusted (not clear to me how so far)
- What happens if the same jte file exists in multiple sources?
- Both gradle and maven plugin expect one root directory for precompiling templates

How are your jte files structured? I'm trying to understand the usecase a bit better, because this feature will complicate things quite a bit and right now I'm unsure if it is worth the effort.

DevAlves1993:
Hello @casid 

> the IntelliJ plugin does not know about this yet and would need to be adjusted (not clear to me how so far)

Using JTE's IntelliJ plugin hasn't caused me any problems, and the autocompletion offered by the latter works correctly, regardless of the directory in which the JTE template is located.

> What happens if the same jte file exists in multiple sources ?

The file will be retrieved from the first source, and this could be specified in the CompositeCodeResolver documentation. The best would be to specify that template should be named differently when using CompositeCodeResolver.

> Both gradle and maven plugin expect one root directory for precompiling templates

I have to admit that this is going to complicate things a bit in the case of precompilation, the necessary work would have to be done on the gradle and maven plugins.

> How are your jte files structured? I'm trying to understand the usecase a bit better, because this feature will complicate things quite a bit and right now I'm unsure if it is worth the effort.

I usually place template in the resources/templates folder. The Spring View Component library makes it easy to create reusable components with JTE. With Spring View Component, it is necessary to place components in the same package as their classes. As a result, I had traditional template (which were not components) located in the resources/templates folder, while components were located in my class packages.


```java
DirectoryCodeResolver directoryCodeResolverTemplate = new DirectoryCodeResolver(
                Path.of("src/main/resources/templates")
        );
DirectoryCodeResolver directoryCodeResolverComponent = new DirectoryCodeResolver(
                Path.of("src/main/java")

 CompositeCodeResolver compositeCodeResolver
                = new CompositeCodeResolver(
                        List.of(directoryCodeResolverTemplate,
                                directoryCodeResolverComponent));
        return TemplateEngine.create(
                compositeCodeResolver,
                Path.of("jte-classes"),
                ContentType.Html,
                this.getClass().getClassLoader()
        );
```

It was this circumstance that prompted me to propose an implementation of CodeResolver that offers the possibility of reading templates from various locations.

casid:
Hmm, but this approach will fail as soon as you have shared templates (like page layouts) that are used by different pages, no?

For instance this directory structure:

```
src/core/layout.jte
src/module1/page1.jte
src/module1/page2.jte
src/module2/page1.jte
.jteroot
```

The jte root directory would be `src` and here the autocompletion in IntelliJ works. For instance, when you are in `src/module1/page1.jte` you can access the shared layout component:

```
@template.core.layout(title = "Hello, this is my title", content = @`
   <p>Here goes my page content</p>
`)
```

If those are in different root directories, that would not be possible and IntelliJ would mark the layout reference as an error.

If I understand the philosophy of Spring ViewComponent correct, the templates are expected to live along their Java classes right?

A structure like this would also be possible, wouldn't it?

```
src/main/java/core/layout.jte
src/main/java/module1/Page1Controller.java
src/main/java/module1/page1.jte
src/main/java/module1/Page2Controller.java
src/main/java/module1/page2.jte
src/main/java/module2/Page1Controller.java
src/main/java/module2/page1.jte
src/main/java/.jteroot
```

Then, the jte root would be located at `src/main/java` and you would not need to mix templates in resources with templates in java packages.

DevAlves1993:
> If those are in different root directories, that would not be possible and IntelliJ would mark the layout reference as an error.

Indeed, it clearly considers it as an error. In my situation, I simply used the class structure generated by "jte" to correctly access the shared layout model, in order to avoid errors.

> If I understand the philosophy of Spring ViewComponent correct, the templates are expected to live along their Java classes right ?

Yes, that's exactly it.

> A structure like this would also be possible, wouldn't it?

Yes, this structure works well.

Thanks @casid

PR #251, opened 2023-07-14
refactored spring starter

tschuehly:
There are some inconsistencies I've cleaned up.

I've also fixed an Issue that you couldn't overwrite the writer of the TemplateEngine, as JteView used the outputStream directly

@atomfrede  Maybe you can take a look at it?

### DIFF
--- a/jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
+++ b/jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
@@ -21,16 +21,15 @@ import java.nio.file.Paths;
 @EnableConfigurationProperties(JteProperties.class)
 public class JteAutoConfiguration {
 
-    @Autowired
-    private Environment environment;
+    private final JteProperties jteProperties;
 
-    @Autowired
-    private JteProperties jteProperties;
+    public JteAutoConfiguration(JteProperties jteProperties) {
+        this.jteProperties = jteProperties;
+    }
 
     @Bean
     @ConditionalOnMissingBean(JteViewResolver.class)
     public JteViewResolver jteViewResolver(TemplateEngine templateEngine) {
-
         return new JteViewResolver(templateEngine, jteProperties.getTemplateSuffix());
     }
 
@@ -39,8 +38,8 @@ public class JteAutoConfiguration {
     @ConditionalOnMissingBean(TemplateEngine.class)
     public TemplateEngine jteTemplateEngine() {
 
-        if (jteProperties.isProductionEnabled(environment)) {
-            // Templates will be compiled by the maven build task
+        if (jteProperties.usePreCompiledTemplates()) {
+            // Templates will need to be compiled by the maven/gradle build task
             return TemplateEngine.createPrecompiled(ContentType.Html);
         } else {
             // Here, a JTE file watcher will recompile the JTE templates upon file save (the web browser will auto-refresh)
--- a/jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
+++ b/jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
@@ -1,7 +1,7 @@
 package gg.jte.springframework.boot.autoconfigure;
 
 import gg.jte.TemplateEngine;
-import gg.jte.output.Utf8ByteOutput;
+import gg.jte.output.PrintWriterOutput;
 import org.springframework.http.MediaType;
 import org.springframework.web.servlet.view.AbstractTemplateView;
 
@@ -24,16 +24,10 @@ public class JteView extends AbstractTemplateView {
 
     @Override
     protected void renderMergedTemplateModel(Map<String, Object> model, javax.servlet.http.HttpServletRequest request, javax.servlet.http.HttpServletResponse response) throws Exception {
-
         String url = this.getUrl();
-
-        Utf8ByteOutput output = new Utf8ByteOutput();
-        templateEngine.render(url, model, output);
-
         response.setContentType(MediaType.TEXT_HTML_VALUE);
         response.setCharacterEncoding(StandardCharsets.UTF_8.name());
-        response.setContentLength(output.getContentLength());
-
-        output.writeTo(response.getOutputStream());
+        PrintWriterOutput output = new PrintWriterOutput(response.getWriter());
+        templateEngine.render(url, model, output);
     }
 }
--- a/jte-spring-boot-starter-3/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
+++ b/jte-spring-boot-starter-3/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
@@ -1,27 +1,28 @@
 package gg.jte.springframework.boot.autoconfigure;
 
-import gg.jte.*;
-import gg.jte.resolve.*;
-import org.springframework.beans.factory.annotation.*;
-import org.springframework.boot.autoconfigure.*;
-import org.springframework.boot.autoconfigure.condition.*;
-import org.springframework.boot.context.properties.*;
-import org.springframework.context.annotation.*;
-import org.springframework.core.env.*;
-import org.springframework.web.servlet.view.*;
+import gg.jte.CodeResolver;
+import gg.jte.ContentType;
+import gg.jte.TemplateEngine;
+import gg.jte.resolve.DirectoryCodeResolver;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.web.servlet.view.AbstractTemplateViewResolver;
 
-import java.nio.file.*;
+import java.nio.file.FileSystems;
+import java.nio.file.Paths;
 
 @AutoConfiguration
 @ConditionalOnClass({TemplateEngine.class, AbstractTemplateViewResolver.class})
 @EnableConfigurationProperties(JteProperties.class)
 public class JteAutoConfiguration {
+    private final JteProperties jteProperties;
 
-    @Autowired
-    private Environment environment;
-
-    @Autowired
-    private JteProperties jteProperties;
+    public JteAutoConfiguration(JteProperties jteProperties) {
+        this.jteProperties = jteProperties;
+    }
 
     @Bean
     @ConditionalOnMissingBean(JteViewResolver.class)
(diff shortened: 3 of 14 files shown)


tschuehly:
I wasn't quite sure how to test the starter as I'm not sure how to do that for my own library yet, maybe with some IntegrationTests and a TestApplication how I did it here: https://github.com/tschuehly/spring-view-component/blob/b751f067f69a780734ff44a7d97247bd888f0e6a/jte/src/test/kotlin/de/tschuehly/spring/viewcomponent/jte/IntegrationTest.kt

tschuehly:
Another question, in the current REAMDE: https://github.com/tschuehly/jte/tree/refactor-spring-starter/jte-spring-boot-starter-3#add-required-dependencies
It tells me you need to include Jte itself as dependency aswell, can't we expose this as an api depencency?

atomfrede:
Looks good. I like the config property meta data thing :D Regarding the jte dependency itself, I think I used the thymeleaf starter as a "template", so maybe lets have a look how its done there and do the same, so there are no suprises. 
Regarding testing, totally right, I used a sample application to test it. 
I always wanted to update this https://github.com/casid/jte-spring-boot-demo to use the starter. We can use the starter and write some webmvc or playwright tests and run them in github pipelines maybe to have some kind of automated tests, wdyt?

atomfrede (approved):

tschuehly:
Yes I was thinking of a similiar thing, inside the pipeline publish to the local maven repo, then checkout the demo/example project. In my case, I included it inside the repo itself then run the webmvc test against it and check that it contains the proper output.

Or we could use a testapplication right inside the directory. But I did not get the jte codersolver to run inside my test: https://github.com/tschuehly/spring-view-component/blob/master/jte/src/test/kotlin/de/tschuehly/spring/viewcomponent/jte/JteTest.kt

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
```diff
@@ -39,8 +38,8 @@ public JteViewResolver jteViewResolver(TemplateEngine templateEngine) {
     @ConditionalOnMissingBean(TemplateEngine.class)
     public TemplateEngine jteTemplateEngine() {
 
-        if (jteProperties.isProductionEnabled(environment)) {
-            // Templates will be compiled by the maven build task
+        if (jteProperties.usePreCompiledTemplates()) {
```
casid, line 41: Is it still possible to have `usePrecompiledTemplates = true` in production and `usePrecompiledTemplates = false` during local development? Because this setting is going to be what 99% of all user will want to have.

From a security point of view, it would be bad to encourage users to run with hot-reloadable templates in production.

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
```diff
@@ -24,16 +24,10 @@ public boolean checkResource(Locale locale) {
...
-
-        Utf8ByteOutput output = new Utf8ByteOutput();
-        templateEngine.render(url, model, output);
-
         response.setContentType(MediaType.TEXT_HTML_VALUE);
         response.setCharacterEncoding(StandardCharsets.UTF_8.name());
-        response.setContentLength(output.getContentLength());
-
-        output.writeTo(response.getOutputStream());
+        PrintWriterOutput output = new PrintWriterOutput(response.getWriter());
```
casid, line 30: Is there a reason for changing this?

Because if we do this, the `templateEngine.setBinaryStaticContent(true);` needs to be removed, otherwise during rendering all pre-encoded byte arrays will be converted to Strings during rendering.

I don't know how Springs `PrintWriter` behaves. But usually a `PrintWriter` flushes its content from time to time. Have you tested what happens if there is a exception during rendering? Does the user then get a half rendered page?

jte-spring-boot-starter-3/src/main/java/gg/jte/springframework/boot/autoconfigure/JteProperties.java
```diff
@@ -1,15 +1,10 @@
...
-import org.springframework.lang.*;
-
-import java.util.*;
+import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties(prefix = "gg.jte")
 public class JteProperties {
-
-    private String productionProfileName = "prod";
+    private Boolean usePrecompiledTemplates = false;
```
casid, line 7: This should never be `null`, so `boolean` would be better.

casid (commented):
@tschuehly sorry for the late reply, I left some comments for review.

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
casid: Ah, I've seen your commit message with the reason behind this change.

`templateEngine.setBinaryStaticContent(true);` should then still be removed, in order to not make the rendering unreasonably slow.

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
tschuehly: Why is it bad?
The best way would probably to do the template resolving at startup in production.

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
tschuehly: Done :)

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
casid: It is bad because it requires a JDK at runtime, not just a JRE. From a security point of view it is better to precompile templates on the build server, instead of the production server. The same way Kotlin/Java sources are compiled on the build server and not on prod.

> Just Files.walk trough the directory and resolve them all on startup?

What do you mean by that?

jte-spring-boot-starter-3/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
```diff
@@ -34,17 +35,15 @@ public JteViewResolver jteViewResolver(TemplateEngine templateEngine) {
     @Bean
     @ConditionalOnMissingBean(TemplateEngine.class)
     public TemplateEngine jteTemplateEngine() {
-        if(jteProperties.usePreCompiledTemplates()){
+        if (Boolean.TRUE.equals(jteProperties.usePreCompiledTemplates())) {
```
casid, line 38: Why are you boxing `boolean` here?

This should just be `if(jteProperties.usePreCompiledTemplates())`

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
tschuehly: Hey casid, 
sorry for the late reply. I think I have different solution that is a bit more work but would remove the necessity of using a build plugin.

I'm currently building an annotation processor for preprocessing of my html templates for Spring ViewComponent. This is primarily for HTMX attributes but I need to compile the JTE Templates aswell.

https://github.com/tschuehly/spring-view-component/blob/v0.6.2-SNAPSHOT/core/src/main/kotlin/de/tschuehly/spring/viewcomponent/core/processor/ViewComponentProcessor.kt

For the JTE Spring Boot Starter I think I could create an annotation processor that targets the @SpringBootApplication Annotation and then compiles the whole template folder at build time into the build/classes for gradle and target/classes for maven

For Hot Reload at development time I would implement a FileChangeListener that recompiles the changed file into the build folder.
https://github.com/tschuehly/spring-view-component/blob/v0.6.2-SNAPSHOT/core/src/main/kotlin/de/tschuehly/spring/viewcomponent/core/processor/ViewComponentChangeListener.kt

I did a bit of digging into the JTE sources and I think I would just create a precompiled TemplateEngine and then use the gg.jte.compiler.TemplateCompiler.precompile() to compile a single template right?

What are your thoughts on that solution @casid?

jte-spring-boot-starter-3/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
tschuehly: SonarLint screamed at me to that, and as I'm normally using Kotlin I thought there must be a good reason to do so

jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteAutoConfiguration.java
tschuehly: @casid 
I was look at the jstachio docs and they also use the annotation processor:

> JStachio does its type checking and code generation with the Java annotation processor which is built into the JDK. JTE and Rocker require pre-compiling with a build plugin. Thus you cannot precompile easily if your build system is not supported. JTE, Rocker, and JSP do not actually read the symbolic tree of the Java models but rather just translate the code to Java.

https://github.com/jstachio/jstachio/blob/main/why_jstachio_is_better.md

casid: I don't understand what the annotation processor should accomplish, as jte doesn't make use of annotations like jstachio does. Wouldn't this just introduce another layer of magic?

Both hot reload and precompiling templates work with the current spring boot starter already, no?

Maybe what you want is more specific to the Spring View Component library rather than to a general purpose spring boot starter plugin?

tschuehly: Well, the annotation processor would solve this problem:
> Is it still possible to have `usePrecompiledTemplates = true` in production and `usePrecompiledTemplates = false` during local development? Because this setting is going to be what 99% of all user will want to have.
> 
> From a security point of view, it would be bad to encourage users to run with hot-reloadable templates in production.

You would need to have just one annotation in your whole project to enable it. With Spring you could just use @SpringBootApplication

The Annotation processor could replace the gradle/maven plugin. 

If you are not open to that idea I would just update the docs, with a reference that you need to use the maven/gradle plugin in production.

casid: This problem is already solved in the current version of the Spring Boot Starter. Maybe we should just keep it this way?

```
if (jteProperties.isProductionEnabled(environment)) {
    // Templates will be compiled by the maven build task
```

Updating the documentation with a hint to maven/gradle would be very good.

tschuehly: IMO this hides the behaviour and makes it less obvious, that's why I changed it. 

It is not guaranteed that in production the profile is "prod", that's why I think with an update to the Readme with:
> In Production you should use the Gradle/Maven Plugin and set usePreCompiledTemplates to true, as this doesn't require a JDK at runtime and is safer

tschuehly: I've adjusted the README and the autoconfiguration. Users will need to explicitly specify development mode or precompiled templates. 
If they do not the application will not start

casid: Nice!

tschuehly:
Let's do it 👍

Issue #254, opened 2023-07-28
Support Dynamic Html Attribute Names


wyaeld:
Hi @casid 

My Team has recently introduced StimulusJs to our Jte/Ktor application, and are finding a problem with the inability to build Html Attribute Names using expressions.

Stimulus is a JS project, part of the Hotwire project that's extracted from RubyOnRails, but isn't commonly found yet in other stacks.
https://stimulus.hotwired.dev/

The core issue, is that Stimulus is a form of declarative controller binding, but uses both the name and value of html attributes to convey instructions, for instance (from their docs)

```html
<div data-controller="hello">
  <input data-hello-target="name" type="text">

  <button data-action="click->hello#greet">
    Greet
  </button>

  <span data-hello-target="output">
  </span>
</div>
```

The `data-hello-target` is a specific binding to the `target` value of the `hello-controller`

We want to be able to build inputs, selects and other elements as Jte templates that can be reused, but supply things like the `controller` name as a variable to be passed in.

Is this something that Jte could gain support for?

casid:
Hi @wyaeld,

that's cool, I wanted to try Hotwire/Stimulus with jte for a while, but haven't got the chance to do this yet. Is your project open source? If so, I'd be curious to have a look at it :-)

As for this issue, you should be able to create a custom `HtmlPolicy` to allow output in attribute names.

```java
// Same as gg.jte.html.OwaspHtmlPolicy, but allows output in tags and attributes
public class MyHtmlPolicy extends PolicyGroup {
    public MyHtmlPolicy() {
        addPolicy(new PreventUppercaseTagsAndAttributes());
        // addPolicy(new PreventOutputInTagsAndAttributes());
        addPolicy(new PreventUnquotedAttributes());
    }
}
```

The policy can be set through `gg.jte.TemplateEngine#setHtmlPolicy`.

For precompiling with maven/gradle plugin the policy class need to be configured there as well, through `htmlPolicyClass` property.

wyaeld:
It's an internal company system, so I can't share, but once we iron out these last kinks, I'm having to write you up a doc example of integrating ktor/jte/hotwire and stimulus.  Its not much work at all, and I think its a nice combination.

On your recommendations.  Doing the above does work for a single attribute expression.  We are running into this issue though

```java
    @Test
    void dynamicAttribute2_for_hotwire() {
        class HotwireHtmlPolicy extends PolicyGroup {
            public HotwireHtmlPolicy() {
[...]

PR #256, opened 2023-08-02
Support Dynamic Html Attribute Names

casid:
For hotwire development it is required to have dynamic attribute names in certain cases. See issue #254

This is prevented by `OwaspHtmlPolicy`, or more specifically the policy `PreventOutputInTagsAndAttributes`. This is because attribute names are not a safe slot to put untrusted user input into. There are various ways to escape the attribute name.

This PR fixes a bug in jte, so that correct output is generated, if this policy is disabled by the user and dynamic output is used in an attribute name.

There are currently two ways to achieve this:
- Custom HTML Policy that disables or weakens the `PreventOutputInTagsAndAttributes` policy
- Using `$unsafe` for the dynamic output in the attribute name

@wyaeld during implementation I noticed that my initial recommendation to disable the `PreventOutputInTagsAndAttributes` policy entirely does remove a few more policies that are quite useful and I think that most users will want to have, such as preventing dynamic tag names and preventing control structures and content blocks in attribute names.

That's why I've parameterized the `PreventOutputInTagsAndAttributes`, to make it possible to only disable the check that you need, like this: https://github.com/casid/jte/blob/66dbec00f9be247e06a200e531b2a3ecdd2a0c08/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngine_HtmlOutputEscapingTest.java#L1419

As an alternative, you could also use `$unsafe` for these parts and stick with the default `OwaspHtmlPolicy`, like here: https://github.com/casid/jte/blob/66dbec00f9be247e06a200e531b2a3ecdd2a0c08/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngine_HtmlOutputEscapingTest.java#L1448

### DIFF
--- a/jte-kotlin/src/main/java/gg/jte/compiler/kotlin/KotlinCodeGenerator.java
+++ b/jte-kotlin/src/main/java/gg/jte/compiler/kotlin/KotlinCodeGenerator.java
@@ -319,7 +319,7 @@ public class KotlinCodeGenerator implements CodeGenerator {
     @Override
     public void onHtmlTagAttributeCodePart(int depth, String codePart, String tagName, String attributeName) {
         writeIndentation(depth);
-        kotlinCode.append("jteOutput.setContext(\"").append(tagName).append("\", \"").append(attributeName).append("\")\n");
+        kotlinCode.append("jteOutput.setContext(\"").append(tagName).append("\", \"").appendEscaped(attributeName).append("\")\n");
 
         writeCodePart(depth, codePart);
 
--- a/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngine_HtmlOutputEscapingTest.java
+++ b/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngine_HtmlOutputEscapingTest.java
@@ -5,11 +5,12 @@ import gg.jte.ContentType;
 import gg.jte.TemplateEngine;
 import gg.jte.TemplateException;
 import gg.jte.html.OwaspHtmlPolicy;
-import gg.jte.html.policy.PreventInlineEventHandlers;
-import gg.jte.html.policy.PreventSingleQuotedAttributes;
+import gg.jte.html.policy.*;
 import gg.jte.output.StringOutput;
 import gg.jte.runtime.TemplateUtils;
 import gg.jte.support.LocalizationSupport;
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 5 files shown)


wyaeld:
Been a few days away from keyboard.  Ran things this morning.

Seems to work pretty well.  I'm usually wary about single boolean switched on contructors, but looking at the code I don't see anything obvious that looks better than what you've done there.

 I was also exploring the ergonomics of what happens when we don't have params for it, to try and avoid having to include if/else conditional chunks in the templates.  My understanding was that attributes without values would get no-oped by JTE, but I think I'm misunderstanding.

What should happen here?  I haven't changed the assert from the base test, but wasn't expecting the attributes to render at all.

```java
@Test
        void attributes_dynamicNameForHotwireWithEmptyDefaultsNoParams() {
            codeResolver.givenCode("template.kte", "@param controller:String = \"\"\n@param target:String=\"\"\n<div data-controller=\"${controller}\">" +
             "\n<input data-${controller}-target=\"${target}\"/></div>");

            templateEngine.render("template.kte", TemplateUtils.toMap(), output);

            //assertThat(output.toString()).isEqualTo("<div data-controller=\"hello\">\n<input data-hello-target=\"name\"/></div>");
            assertThat(output.toString()).isEqualTo("<div>\n<input/></div>");
        }
```

This result
```
org.opentest4j.AssertionFailedError: 
expected: 
  "<div>
  <input/></div>"
 but was: 
  "<div data-controller="">
  <input data--target=""/></div>"
```

wyaeld:
Rereading https://github.com/casid/jte/blob/main/DOCUMENTATION.md#smart-attributes

If I made the default values `null` instead of `""` would it work?  Will test

casid:
@wyaeld nope, smart attributes are not working atm when attributes names are dynamic.

This is because smart attributes currently check at compile time if the attribute is a boolean or regular html attribute. In case of boolean it generates different code. That's not possible when the attribute name is not present at compile time.

What's your usecase exactly? Is the controller variable expected to be null, or the target variable?

wyaeld:
Just exploring different ways to organise our templates, and whether we need to have hotwire and non-hotwire variants of common controls, like inputs, selects etc.

What you've done with this PR gives us good building blocks though, its a big improvement.

casid:
Okay, I'll then merge this branch so that it is available in the next bugfix release.

We can still try to add more features later, if needed.

PR #259, opened 2023-08-13
Add JTE support link for http4k

daviddenton:
We have added JTE support to http4k (which was incredibly easy for us - well done! 😄), starting with version 5.6.4.0, and would like the world to know, so this link would be great to add here!


casid:
Thanks for the heads up and for the PR. Glad the integration with http4k worked so smoothly!

PR #265, opened 2023-08-29
bump kotlin ver

Rattlyy:
#264

PR #268, opened 2023-09-02
Bump Kotlin from 1.8.10 to latest 1.9.10.

nedtwigg:
Replacement for

- #265

Looking forward to `3.0.4` with the JTE/KTE fixes, let me know if there are any tasks I can accomplish with PRs which will speed it along.


casid:
Thanks! With this change and the spring boot changes I'll release a `3.1.0` today.

PR #282, opened 2023-10-05
jte-models: Add support to generate Kotlin code

marcospereira:
## What?

This adds support to generate Kotlin code (instead of Java) when using the `jte-models` extension. There are a few inconveniences when mixing both languages (clashing names such as `List`, which are different for both, not having to import standard-lib classes in Kotlin that would break the Java generate code), and having this will solve that.

## How

Simply by providing a `language` configuration to the extension. The default is Java, so people upgrading won't have to change anything in their build files. An example:

```gradle
jte {
    generate()
    binaryStaticContent = true
    jteExtension('gg.jte.models.generator.ModelExtension') {
        language = 'kotlin'
    }
}
```

## Generate code

The generated code looks like this:

```kotlin
@file:Suppress("ktlint")
package test.mytemplates

import gg.jte.models.runtime.*

interface Templates {
    
    @JteView("hello.jte")
    fun hello(message: java.lang.String): JteModel

}
```

```kotlin

@file:Suppress("ktlint")
package test.mytemplates

import gg.jte.TemplateEngine
import gg.jte.models.runtime.*

class DynamicTemplates(private val engine: TemplateEngine) : Templates {
    
    override fun hello(message: java.lang.String): JteModel {
        val paramMap = mapOf(
        
            "message" to message,
        )
        return DynamicJteModel(engine, "hello.jte", paramMap);
    }

}
```

```kotlin

@file:Suppress("ktlint")
package test.mytemplates

import gg.jte.models.runtime.*
import gg.jte.ContentType
import gg.jte.TemplateOutput
import gg.jte.html.HtmlInterceptor
import gg.jte.html.HtmlTemplateOutput

class StaticTemplates : Templates {
    
    override fun hello(message: java.lang.String): JteModel {
        return StaticJteModel<TemplateOutput>(
            ContentType.Plain,
            { output, interceptor -> test.mytemplates.JtehelloGenerated.render(output, interceptor, message) },
            "hello.jte",
            "test.mytemplates",
            test.mytemplates.JtehelloGenerated.JTE_LINE_INFO
        );
    }

}
```

### DIFF
--- a/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
+++ b/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
@@ -18,6 +18,17 @@ public class ModelConfig {
         return map.getOrDefault("implementationAnnotation", "");
     }
 
+    public Language language() {
+        String configuredLanguage = map.getOrDefault("language", "JAVA").toUpperCase();
+        Language language = Language.JAVA;
+        try {
+            language = Language.valueOf(configuredLanguage);
+        } catch (IllegalArgumentException ex) {
+            // how to report wrong conciguration? fail or default to Java?
+        }
+        return language;
+    }
+
     public Pattern includePattern() {
         String includePattern = map.get("includePattern");
         if (includePattern == null) {
--- a/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
+++ b/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
@@ -22,7 +22,7 @@ public class ModelConfig {
     public Language language() {
         String configuredLanguage = map.getOrDefault("language", "Java");
         try {
-            return Language.valueOf(configuredLanguage.toUpperCase());
+            return Language.valueOf(configuredLanguage);
         } catch (IllegalArgumentException ex) {
             String supportedValues = Arrays.toString(Language.values());
             throw new IllegalArgumentException(
--- a/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
+++ b/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
@@ -20,7 +20,7 @@ public class ModelConfig {
     }
 
     public Language language() {
-        String configuredLanguage = map.getOrDefault("language", "JAVA");
+        String configuredLanguage = map.getOrDefault("language", "Java");
         try {
             return Language.valueOf(configuredLanguage.toUpperCase());
         } catch (IllegalArgumentException ex) {
--- a/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
+++ b/jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
@@ -1,5 +1,6 @@
 package gg.jte.models.generator;
 
+import java.util.Arrays;
 import java.util.Map;
 import java.util.regex.Pattern;
 
@@ -19,14 +20,16 @@ public class ModelConfig {
     }
 
     public Language language() {
-        String configuredLanguage = map.getOrDefault("language", "JAVA").toUpperCase();
-        Language language = Language.JAVA;
+        String configuredLanguage = map.getOrDefault("language", "JAVA");
         try {
-            language = Language.valueOf(configuredLanguage);
+            return Language.valueOf(configuredLanguage.toUpperCase());
         } catch (IllegalArgumentException ex) {
-            // how to report wrong conciguration? fail or default to Java?
+            String supportedValues = Arrays.toString(Language.values());
+            throw new IllegalArgumentException(
+                String.format("JTE ModelExtension 'language' property is not configured correctly (current value is '%s'). Supported values: %s", configuredLanguage, supportedValues),
+                ex
+            );
         }
-        return language;
     }
 
     public Pattern includePattern() {
--- a/jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
+++ b/jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
@@ -8,7 +8,6 @@ import gg.jte.extension.api.TemplateDescription;
 
 import java.nio.file.Path;
 import java.util.Collection;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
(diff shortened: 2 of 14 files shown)


jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
```diff
@@ -18,6 +18,17 @@ public String implementationAnnotation() {
         return map.getOrDefault("implementationAnnotation", "");
     }
 
+    public Language language() {
+        String configuredLanguage = map.getOrDefault("language", "JAVA").toUpperCase();
+        Language language = Language.JAVA;
+        try {
+            language = Language.valueOf(configuredLanguage);
+        } catch (IllegalArgumentException ex) {
+            // how to report wrong conciguration? fail or default to Java?
```
marcospereira, line 27: @casid, I'm unsure how (or even if) to do this. It is an open question for this PR.

jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
```diff
@@ -21,7 +21,7 @@ public class ModelExtension implements JteExtension {
 
     @Override
     public String name() {
-        return "Generate type-safe model facade for templates";
+        return "Generate type-safe model facade for templates in Java";
```
marcospereira, line 24: I'm also unsure about the impact of changing this.

jte-models/src/main/java/gg/jte/models/generator/ModelGenerator.java
```diff
@@ -23,17 +23,27 @@ public class ModelGenerator {
     private final String targetClassName;
     private final String interfaceName;
 
+    private final Language language;
+
     public ModelGenerator(TemplateEngine engine, String templateSubDirectory, String targetClassName, String interfaceName) {
```
marcospereira, line 28: I kept this constructor for compatibility reasons. But it may be unnecessary now since it is not used in the code base.

marcospereira (commented):
A few comments and questions.

jte-models/src/main/java/gg/jte/models/generator/ModelGenerator.java
edward3h: ModelGenerator is internal implementation, so it is not necessary to keep the constructor overload.

jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
edward3h: In my opinion you should fail the build in this case. Building Java when a different language was specified is not what the user wants.

I think throwing an exception with a good description would be good enough, but you should test that.

jte-models/src/main/java/gg/jte/models/generator/ModelExtension.java
edward3h: The name isn't actually used anywhere at the moment.

edward3h:
Looks good to me. Just one question: since Kotlin can interoperate with Java classes, what is the benefit of generating Kotlin versions of the classes? (I'm not much of a Kotlin user)

marcospereira:
> Looks good to me. Just one question: since Kotlin can interoperate with Java classes, what is the benefit of generating Kotlin versions of the classes? (I'm not much of a Kotlin user)

Interoperability is quite good, but there are places where it fails. For example, if you have a Kte template such as:

```html
@param messages: List<String>

<html lang="pt">
    <head>
        <title>First Page</title>
    </head>
    <body>
        <h1>message</h1>
    </body>
</html>
```

`List` is a `kotlin.collections.List`, and users don't need to import manually. The generated (Java) interface misses the import and is then broken:

```java
package br.ufpe.liber;
import gg.jte.models.runtime.*;

public interface Templates {
    
    @JteView("index.kte")
    JteModel index(List<String> messages);

}
```

Compilation error:

```
Templates.java:8: error: cannot find symbol
    JteModel index(List<String> messages);
                   ^
  symbol:   class List
  location: interface Templates
```

There are workarounds, but then the user needs to use them constantly, which impacts the developer experience.

jte-models/src/main/java/gg/jte/models/generator/ModelConfig.java
marcospereira: Done in 440c1f0.

jte-models/src/main/java/gg/jte/models/generator/ModelGenerator.java
marcospereira: Removed in aac2a43.

marcospereira:
Thanks for reviewing, @edward3h. I pushed a couple of commits after your feedback. :-)

edward3h:
Looks good. Thank you for your contribution.

casid:
Thanks for the contribution @marcospereira and thanks for the review @edward3h.

I know it is against the Java conventions, but so far enum constants in the jte project have been camel case. Could we rename the Language enum values to `Java` and `Kotlin`?

I think it is also okay to be case sensitive here when we parse the setting and do a `Language.valueOf(configuredLanguage);` as you had it initially. This will always be an error message at build time, never at runtime. We probably do not want to support users entering `jAvA` and the like :-)

Other than that, this looks fantastic!

marcospereira:
Thanks, @casid. 



> I know it is against the Java conventions, but so far enum constants in the jte project have been camel case. Could we rename the Language enum values to `Java` and `Kotlin`?

Done in a059cd3.



> I think it is also okay to be case sensitive here when we parse the setting and do a `Language.valueOf(configuredLanguage);` as you had it initially. This will always be an error message at build time, never at runtime. We probably do not want to support users entering `jAvA` and the like :-)

Done in 1888f90. 

This should be good to go now. :-)

casid:
Wow, that was quick! I'll merge this as soon as the CI build is done.

marcospereira:
> Wow, that was quick! I'll merge this as soon as the CI build is done.

Thanks! Also, when can we have a release with those changes?

casid:
There haven't been any changes since the last release, so there's nothing blocking us from releasing this asap. Will have a look at it this weekend.

marcospereira:
> There haven't been any changes since the last release, so there's nothing blocking us from releasing this asap. Will have a look at it this weekend.

Thank you. No rush here, though. :-)

By the way, I just submitted a follow-up pr to update the docs. See #283. :-)

casid:
@marcospereira I published version 3.1.2 which contains this feature!

PR #283, opened 2023-10-06
jte-models: Update docs with correct language values

marcospereira:
## What?

This is a quick follow up to #282. After changing the configuration to be case
sensitive, the documentation examples got outdated (ooops!)
.


casid:
No worries! Thanks for spotting this and the quick fix.

PR #285, opened 2023-10-10
KTE: Generate templates using default parameter values

marcospereira:
## What?

Kotlin supports default parameters at method definition. For example:

```kotlin
fun hello(name: String = "jte"): JteModel
```

But those are not used when generating code. This adds the default values, except for `gg.jte.Content` since the syntax (<code>@``</code>) is not supported in Kotlin Code. This will allow users to call templates from Kotlin code without the need to pass the already defined (at the templates) default values.

Currently, only calls from another template recognize the default values.


jte-kotlin/src/main/java/gg/jte/compiler/kotlin/KotlinCodeGenerator.java
```diff
@@ -73,6 +73,13 @@ public void onParam(String parameter) {
         }
         kotlinCode.append(paramInfo.name).append(':').append(paramInfo.type);
 
+        // If there is a default value that IS NOT a gg.jte.Content, then add it to the
+        // method definition. gg.jte.Content are excluded because the syntax is not supported
+        // by Kotlin.
+        if (paramInfo.defaultValue != null && !paramInfo.defaultValue.startsWith("@`")) {
```
marcospereira, line 79: The same logic repeats here:

https://github.com/casid/jte/pull/285/files#diff-8bf3a3fda545029327a72c0eb5187d04be7a81260716bb2b04a31eecc3ed2029R20

Which I'm not happy with, but I didn't want to pollute `ParaInfo` with code generation logic. Maybe have it declared somewhere where it can be reused?

jte-kotlin/src/test/java/gg/jte/compiler/kotlin/TestKotlinParamInfo.java
```diff
@@ -0,0 +1,103 @@
+package gg.jte.compiler.kotlin;
+
+import gg.jte.compiler.ParamInfo;
+import gg.jte.compiler.TemplateParserVisitor;
+import gg.jte.compiler.TemplateParserVisitorAdapter;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class TestKotlinParamInfo {
```
marcospereira, line 9: There is some repetition in the tests here, but I prefer it that way since they are easier to read (IMO). Let me know if you want me to refactor it somehow to better match the project's style.

jte-models/src/main/jte/dynamictemplates/kmethod.jte
```diff
@@ -4,7 +4,7 @@
 @param TemplateDescription template
 
     override fun ${Util.methodName(template)}(${Util.kotlinTypedParams(template)}): JteModel {
-        val paramMap = mapOf(
+        val paramMap = mapOf<String, Any?>(
```
marcospereira, line 7: This is a minor fix after #282. When the map is empty, the type cannot be inferred. Having it explicitly declared here works for all the cases.

jte-models/src/test/java/gg/jte/models/generator/TestModelExtension.java
```diff
@@ -14,75 +17,232 @@
...
 
 import static gg.jte.extension.api.mocks.MockConfig.mockConfig;
+import static gg.jte.extension.api.mocks.MockParamDescription.mockParamDescription;
 import static gg.jte.extension.api.mocks.MockTemplateDescription.mockTemplateDescription;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
 
+@DisplayName("jte-models Extension tests")
 public class TestModelExtension {
```
marcospereira, line 29: I heavily refactored these tests, but I think they are easier to read now, and a few test cases were added. Again, some repetition.

marcospereira (commented):
Notes for reviewers. :-)

casid (commented):
Thanks for the PR, I assume this will be quite useful for Kotlin users! I left a few notes about the tests, I think we should have two more black box tests. Besides that, this looks good to me.

jte-kotlin/src/test/java/gg/jte/compiler/kotlin/TestKotlinParamInfo.java
marcospereira: Got it. So, this PR introduces a test for _regular_ (non-Content) params with default values:

https://github.com/casid/jte/pull/285/files#diff-b16684272d3912ace7de14f3b2499de85c89003e7ee137ebd4af72bf89e41fc2R721

The two last points were addressed by 3e708872665f6ece25634e8e368c3d8ae18e6ff8.

jte-models/src/main/java/gg/jte/models/generator/Util.java
```diff
@@ -14,8 +13,15 @@ public static String typedParams(TemplateDescription template) {
         return template.params().stream().map(param -> String.format("%s %s", param.type(), param.name())).collect(Collectors.joining(", "));
     }
 
-    public static String kotlinTypedParams(TemplateDescription template) {
-        return template.params().stream().map(param -> String.format("%s: %s", param.name(), param.type())).collect(Collectors.joining(", "));
+    public static String kotlinTypedParams(TemplateDescription template, boolean includeParamDefaultValue) {
+        return template.params().stream().map(param -> {
+            StringBuilder formatedParam = new StringBuilder();
```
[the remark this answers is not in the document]
marcospereira: Fixed in 45ca217b1d92f99394c69f0a6e08d5fb425bda47.

jte-kotlin/src/test/java/gg/jte/compiler/kotlin/TestKotlinParamInfo.java
marcospereira: Hum, got a few unexpected test failures:

https://github.com/casid/jte/actions/runs/6485276269/job/17611002039?pr=285#step:5:312

@casid, I can investigate those later, but if you know what those are about, that can speed up my investigation.

marcospereira: So, considering this test:

```java
    @Test
    void tagWithDefaultContentParam() {
        givenTag("tagWithDefaultContentParam", "@param one:Int = 1\n" +
                          "@param content:gg.jte.Content = @`Some Content`\n" +
                          "${one}, ${content}");
        givenTemplate("@template.tag.tagWithDefaultContentParam()");

        thenOutputIs("1, Some Content");
    }
```

The code generated for the tag when running `gg.jte.kotlin.TemplateEngine_BinaryTest` was:

```kotlin
@file:Suppress("ktlint")
package gg.jte.generated.ondemand.tag
@Suppress("UNCHECKED_CAST", "UNUSED_PARAMETER")
class JtetagWithDefaultContentParamGenerated {
    companion object {
    	@JvmField val JTE_NAME = "tag/tagWithDefaultContentParam.kte"
    	@JvmField val JTE_LINE_INFO = intArrayOf(0,0,0,0,0,0,0,0,0,0,2,2,2,2,2)
    	@JvmStatic val BINARY_CONTENT = gg.jte.runtime.BinaryContent.load(JtetagWithDefaultContentParamGenerated.javaClass, "JtetagWithDefaultContentParamGenerated.bin", 14,18);
    	@JvmStatic val TEXT_PART_BINARY_0 = BINARY_CONTENT.get(0)
    	@JvmStatic val TEXT_PART_BINARY_1 = BINARY_CONTENT.get(1)
    	@JvmStatic fun render(jteOutput:gg.jte.TemplateOutput, jteHtmlInterceptor:gg.jte.html.HtmlInterceptor?, one:Int = 1, content:gg.jte.Content) {
    		jteOutput.writeBinaryContent(TEXT_PART_BINARY_0)
    		jteOutput.writeUserContent(one)
    		jteOutput.writeBinaryContent(TEXT_PART_BINARY_1)
    		jteOutput.writeUserContent(content)
    	}
    	@JvmStatic fun renderMap(jteOutput:gg.jte.TemplateOutput, jteHtmlInterceptor:gg.jte.html.HtmlInterceptor?, params:Map<String, Any?>) {
    		val one = params["one"] as Int? ?: 1
    		val content = params["content"] as gg.jte.Content? ?: object : gg.jte.Content {
    			override fun writeTo(jteOutput:gg.jte.TemplateOutput) {
    				jteOutput.writeBinaryContent(TEXT_PART_BINARY_2)
    			}
    		}
    		render(jteOutput, jteHtmlInterceptor, one, content);
    	}
    }
}
```

The culprit is that line `jteOutput.writeBinaryContent(TEXT_PART_BINARY_2)` since there is only `TEXT_PART_BINARY_0` and `TEXT_PART_BINARY_1`. If I got it correctly, that code is generated here:

https://github.com/casid/jte/blob/e700eb01fe7272744e302075fc098b173cc66c77/jte-kotlin/src/main/java/gg/jte/compiler/kotlin/KotlinCodeGenerator.java#L254-L263

I'm not sure it should have been `binaryTextParts.size()` if `TEXT_PART_BINARY_` variables are 0-based. 🤔 

@casid, what do you think? Am I barking at the wrong tree?

marcospereira: Hum, yeah, this seems bug. Also, I couldn't find a single test that mixes:

1. Binary rendering
2. Non-primitive parameters with non-null default values

These two tests:

https://github.com/casid/jte/blob/e700eb01fe7272744e302075fc098b173cc66c77/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngine_HtmlOutputEscapingTest.java#L1003-L1019

They are almost there, but they are not running with `templateEngine.setBinaryStaticContent(true)`. I then created a subclass like this:

```java
package gg.jte.kotlin;

import org.junit.jupiter.api.BeforeEach;

public class TemplateEngine_HtmlOutputEscapingBinaryTest extends TemplateEngine_HtmlOutputEscapingTest {

    @BeforeEach
    void setUp() {
        templateEngine.setBinaryStaticContent(true);
    }
}
```

And... both tests fail. I don't have the bandwidth right now to fix the problem, but at least this investigation should point you to the right direction.

casid: I think you found a bug with default content parameters and binary templates. I'll try to have a look at it this weekend!

casid: @marcospereira thanks for finding this and for your investigation! I just pushed a fix for this to the main branch.

casid: Huh, this change fixed another bug, where line numbers where incorrectly displayed as 0 instead of the real one. Pushed another commit to adjust the tests for this.

marcospereira:
Thanks, @casid!

I've rebased against the main branch, and this should be good to go. :-)

casid:
Yes! Thank you for the contribution, merged and closed!

marcospereira:
> Yes! Thank you for the contribution, merged and closed!

Thank you. Contributing to the project has been a great experience. :-)

Do you have an idea of when a new release will be made?

PR #287, opened 2023-10-12
ci: Do not run code coverage on forks

marcospereira:
## What?

Code coverage job will fail on forks since they cannot upload results to
Codecov. This adds a condition to control if the job should run or not. For
more details see GitHub Actions docs:

https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution


casid:
Thanks for the PR, that's very useful!

PR #288, opened 2023-10-13
Use MkDocs for project documentation

marcospereira:
## What?

tl;dr

Use [MkDocs](https://www.mkdocs.org/), [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/), and [mike](https://github.com/jimporter/mike) to write, publish, and version the docs.

## Current state

See <https://marcospereira.github.io/jte/>.

## How it works

I will add plenty of links here so that it can function as a reference later.

### Overall

The docs are now [searchable](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/?h=search), ~~[versioned](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/)~~, offer [light/dark modes](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#color-palette-toggle), present [code snippets](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#grouping-code-blocks) for Java and Kotlin, have [fast navigation](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/?h=instant#instant-loading), automatically generated [table of contents](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown/#table-of-contents), and a few other markup extensions.

It is [Python Markdown](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown/) with a few [extensions](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/) enabled. Although there are differences between that and [GFM](https://github.github.com/gfm/), they are minor, and I thought it was easy to adapt, considering the live-edit experience offered by MkDocs.

### Locally

Python 3 is a requirement to install MkDocs and its plugins. The packages to install (at the time of this writing) are:

- [mkdocs-material](https://pypi.org/project/mkdocs-material/): `pip install mkdocs-material`
- [mkdocs-awesome-pages-plugin](https://pypi.org/project/mkdocs-awesome-pages-plugin/): `pip install mkdocs-awesome-pages-plugin`
- ~~[mkdocs-macros-plugin](https://pypi.org/project/mkdocs-macros-plugin/): `pip install mkdocs-macros-plugin`~~
- [mkdocs-pom-parser-plugin](https://pypi.org/project/mkdocs-pom-parser-plugin/): `pip install mkdocs-pom-parser-plugin`
- ~~[mike](https://pypi.org/project/mike/): `pip install mike`~~

Then, to run the docs locally, execute the following command:

```shell
mkdocs serve
```

This starts a server at <http://localhost:8000/>, and automatically watches and reloads the docs (the browser) as they change.

### Publishing

There is a new workflow to publish the docs automatically on tag creation (for example, when publishing a new release). The workflow only runs for version tags:

```yaml
on:
  push:
    tags:
      - '[0-9]+.[0-9]+.[0-9]+'
```

But it does everything. :-)

For example:

1. Manual step 1: [Publish a release](https://github.com/marcospereira/jte/releases/tag/3.1.3)
2. Automatic step 2: [step 1 triggers a workflow run](https://github.com/marcospereira/jte/actions/runs/6511688476) that pushes to `gh-pages` branch
3. Automatic step 3: [step 2 triggers a gh-pages deployment](https://github.com/marcospereira/jte/actions/runs/6511694948).

## About [`mkdocs-macros`](https://mkdocs-macros-plugin.readthedocs.io/en/latest/)

~~The mkdocs-macro plugins allow us to extend the markup with macros specific to the project. I'm reluctant to include it since the macros are written in Python, but it should be rare that one needs to edit it.~~

The decision was to remove the mkdocs-macros plugin. :-)

## About [mkdocs-pom-parser-plugin](https://pypi.org/project/mkdocs-pom-parser-plugin/)

This is quite useful and allows to reference the project version (defined in the `pom.xml` file) so that there is no need to edit the docs when publishing a new release. For example, we can have links such as:

```markdown
Migrate from [JSP](https://github.com/casid/jte/tree/{{ POM_VERSION }}/jte-jsp-converter)
```

There are other ways to get the version, but I think this is more native to MkDocs.

## Open questions/tasks?

- [x] Remove the macros plugin?
- [ ] Configure Github Pages. See [this comment](https://github.com/casid/jte/pull/288#issuecomment-1769877310) for details.


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,47 @@
+name: Publish Docs
+
+on:
+  push:
+    tags:
+      # Only runs for version tags:
+      # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs
+      - '[0-9]+.[0-9]+.[0-9]+'
```
marcospereira, line 8: We can relax this if tags are only created for releases.

docs/binary-rendering.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Binary rendering for max throughput
+
+{{ since("1.7.0") }}
+
+Most template parts are static content, and only a few parts of a template are dynamic. Encoding those static parts repeatedly on every request is wasteful if your web framework sends binary UTF-8 content to the user. jte makes it is possible to encode those static parts at compile time:
+
+=== "Java"
+
+    ```java linenums="1"
```
marcospereira, line 14: For reference, tabbed code blocks need to be indented.

docs/binary-rendering.md
```diff
@@ -0,0 +1,72 @@
+---
+title: jte Binary Rendering
+description: How to use binary rendering for max throughput
```
marcospereira, line 3: These will translate to HTML meta tags:

```
<meta name="description" content="How to use binary rendering for max throughput">
```

marcospereira (commented):
A few extra notes.

.github/workflows/graalvm.yml
```diff
@@ -1,6 +1,19 @@
...
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'mkdocs.yml'
+      - 'main.py'
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - 'mkdocs.yml'
+      - 'main.py'
```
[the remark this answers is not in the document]
casid: This way the regular build pipeline isn't triggered if we only change stuff in the docs, right? Oh, that is very useful! IMHO repetition is minimal and not really a big problem :-)

.github/workflows/publish-docs.yml
casid: So far tags are only created for releases, but I like that you made it very specific here :-)

docs/binary-rendering.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+# Binary rendering for max throughput
+
+{{ since("1.7.0") }}
```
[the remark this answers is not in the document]
casid: Understood. I'm not experienced with the tooling. But if you say the macros bring marginal value to the table and might be a burden to maintain, I'd say let's exclude them for the start.

docs/binary-rendering.md
casid: That's very convenient!

mkdocs.yml
```diff
@@ -0,0 +1,77 @@
...
+  name: material
+  features:
+    - navigation.top
+    - navigation.footer
+    - navigation.instant
+    - search.highlight
+    - content.code.copy
+    - content.tabs.link
+    - content.code.select
+  palette:
```
casid, line 18: Could we set the primary and accent color here? I tested a little bit over here https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/ and teal looks very close to the jte brand color. I looks good and readable in both dark and light mode.

```
primary: teal
accent: teal
```

mkdocs.yml
```diff
@@ -0,0 +1,77 @@
+site_name: jte documentation
+site_description: "Documentation for jte (Java Template Engine)"
+repo_url: https://github.com/casid/jte/
```
casid, line 3: This will become the official jte website, not just the documentation.

I'd suggest to change this to 

```
site_name: jte
site_description: "Secure and speedy templates for Java and Kotlin"
```

DOCUMENTATION.md
casid, line 1: There are some links do this file in the code base itself, so it's very good that we keep this and point to the new documentation.

I can change those once this is merged and jte.gg points to the new site.

casid (commented):
As I said before, this looks amazing. Thank you very much for all your work on this!

I left some minor comments about some wordings and stylings.

About the macros plugin, I'd leave the final decision up to you, since you clearly have more experience with the tooling than I do.

Other than that, this looks good to me!

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,47 @@
...
+
+permissions:
+  contents: write
```
kelunik, line 11: Where does this write to? The current repository? I'd strongly recommend to avoid giving the build pipeline write access to the source code.

kelunik (changes requested):

.github/workflows/graalvm.yml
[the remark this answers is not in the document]
marcospereira: Exactly. No need to run many build steps for simple typo fixes, for example. :-)

docs/binary-rendering.md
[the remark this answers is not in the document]
marcospereira: Removed in f30bd49.

mkdocs.yml
marcospereira: Done in fa0f4f7.

DOCUMENTATION.md
marcospereira: I've went ahead and updated those. Also moved spring-boot-starters docs from the README.md files to the new docs. See 309bede and 9e684bc.

mkdocs.yml
marcospereira: See 6d56a66.

mkdocs.yml
casid: Thanks!

DOCUMENTATION.md
casid: Oh wow, thank you very much!

marcospereira:
Here are a few updates on this:

I've moved the contents in the modules' README.md files to the docs. Because of that, I thought creating a few sections for the docs would be good. The current state is like this:

https://github.com/casid/jte/assets/4576/a1b5bba3-6840-4c1f-b423-b4ba3011aa9b

It is quite easy to shuffle things around, though. See https://github.com/casid/jte/pull/288/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R76-R95.

.github/workflows/publish-docs.yml
marcospereira: > Where does this write to?

Precisely for this workflow, it would write to `gh-pages` branch. I'm not happy with how extensive these permissions are either, but GitHub does not have a way to configure permissions as fine-grained as "push to gh-pages branch". Or at least I couldn't find how to do it. 

I'm not worried about this that much, though. This is a workflow running inside GitHub, not a third-party service. The workflow only changes when someone authorized to write to the `main` branch merges/pushes commits there. Additionally, the workflow is configured to only run for tags (in other words, when someone authorized pushes a new one).

Overall, I think automating this when tags are created makes it worth having such permission. One change I can make though is to move the permissions to the job (instead of being workflow-wide).

What do you think, @kelunik?

marcospereira: > One change I can make though is to move the permissions to the job (instead of being workflow-wide).

See 88cfc4026ea00a51218d4592c9237f7fab3bb89a.

casid: Would it help to grant write permissions to a different repository, for instance jte-website, to publish the data?

marcospereira: > Would it help to grant write permissions to a different repository, for instance jte-website, to publish the data?

It may be tricky to achieve this if the docs source lives here (the ideal place for them to be, IMO). But I see GitHub has a new (beta) feature to publish GH Pages from a workflow, and that would require fewer permissions.

I will try to find the time this week and see how that works. 

But, if this is blocking the PR, I would instead remove it for now. It is here more for the convenience of automatically publishing the docs whenever a release is made.

WDYT?

casid: @marcospereira I think we don't need to rush it. Having this as a separate workflow sounds like a good idea, because then we wouldn't have a branch with lots of changes compared to main and we would not need to give the tooling permission to alter this repository.

This would mean, that we wouldn't need mike at all right? And probably also have no version history?

marcospereira: Sorry, @casid. I'm not sure I got what you said.

Do you mean remove the docs publishing workflow, versioning, and then have the publishing process as a manual step? If so, I'm okay with it too. And yes, that would mean `mike` wouldn't be necessary (which, tbh, simplifies things).

marcospereira:
I'm quickly converting this to a draft so that I can try a new version of the publishing workflow. :-)

.github/workflows/publish-docs.yml
casid: Yes, I think this would mean to remove versioning and mike, which would simplify things.

I don't think we have to have a manual publishing process, I _think_ that it should be possible to adjust the current workflow to use this: https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages 

Looking at the workflow file I think it is already almost 90% done this way. I think we would only remove the ability to write contents in the jte repository, remove mike and pushing to a branch. Sorry in case I got this all wrong!

marcospereira: You got it right. I was looking at the same docs you linked, and I'm adjusting the workflow now. I'm trying it on my fork first, hence moving the PR to draft for now. 🏃🏼‍♂️

casid: I saw your commit after posting my comment :-)

```
permissions:
  # Checkout the project's code
  contents: read
  # Allow to deploy to gh-pages
  pages: write
```

This looks like the tools are only allowed to push to the gh-pages branch? In that case we could keep mike and versioning, right?

marcospereira: > This looks like the tools are only allowed to push to the gh-pages branch? In that case we could keep mike and versioning, right?

No. This means the actions are allowed to upload artifacts to GitHub Pages. But no write access to branches at all (including gh-pages). See my last commit.

casid: Ah, right. Thanks for giving it a try! Let's start with the simplified version. We can still add mike and versioning the day GitHub has more fine-grained permissions.

Let me know when you think this is ready for merging!

marcospereira: Sure! For now, it looks like GH is not super happy about triggering the flow when a new tag is pushed:

https://github.com/marcospereira/jte/actions/runs/6569712256

🤨

marcospereira:
Alright, @casid. We got a winner!

https://github.com/marcospereira/jte/actions/runs/6569837520

Some tricky things to consider (it may be due to me messing around with GH Pages):

1. You have to configure Pages to deploy from a GitHub Actions workflow:
    Screenshot 2023-10-19 at 12 31 09 AM
2. In the repository "Settings -> Environments", there should be a `github-pages`:
    Screenshot 2023-10-19 at 12 32 18 AM
3. In my case, I had to add a "deployment tag rule" to allow tags in that "Environment":
    Screenshot 2023-10-19 at 12 33 51 AM


After that, I created a new release, and everything worked as expected. :-)

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,67 @@
+name: Publish Docs
+
+on:
+  # Runs whenever there is a release/tag creation
+  release:
+    types: [ "published" ]
+
+  #  push:
+  #    tags:
+  #      - '[0-9]+.[0-9]+.[0-9]+'
```
marcospereira, line 10: @casid I don't think it makes a difference if we filter for tags or published releases. Which one do you prefer? I think `release -> published` is more explicit.

casid: Yeah, that's a lot clearer!

casid:
Awesome! I'm heading off to work now, but will merge this in the evening and will try to configure everything as needed to replace the current content no https://jte.gg.

casid (approved):

casid:
Sorry, it got a bit late yesterday.

@marcospereira I merged today and thanks to your instructions I got everything up and running. The new jte website is now live at https://jte.gg/

Thank you so much for your contribution! This is amazing and the documentation is now so much easier to consume than the big file before!

marcospereira:
> @marcospereira I merged today and thanks to your instructions I got everything up and running. The new jte website is now live at https://jte.gg/

Amazing! And thanks for the fixes after merging. 

One thing that I noticed is that since the site was published from the main branch, it suggests a snapshot version here:

https://jte.gg/#getting-started

And that version is used in other places too. I will check later if there is a way to use the latest release/tag instead of the pom version. There are a few other minor things I would like to change/add, so maybe I can bundle them all together. :-)

marcospereira:
> One thing that I noticed is that since the site was published from the main branch

Oh, and about this, cutting a new release would be good to both validate the workflow, and also have #285 out of the door. But it is Friday, so no rush at all. :-)

PR #289, opened 2023-10-20
docs: Use latest git tag instead of pom version

marcospereira:
As commented here:

https://github.com/casid/jte/pull/288#issuecomment-1772720971

The current docs use the version defined in the parent pom.xml file, which shows a snapshot version when publishing from the main branch. This adds a MkDocs plugin to make it possible to use the latest tag instead.


casid:
Thanks for the fix!

PR #290, opened 2023-10-21
Fix broken links

marcospereira:
## What?

At some point during #288, I replaced old links, adding a `/latest/` (since the docs were supposed to be versioned). This fixes those broken links and adds a new workflow to check links across the project. A few other broken links were fixed during the process.

For reference, it uses https://github.com/lycheeverse/lychee.


casid (approved):

casid:
Thank you for the fix!

casid:
Lychee looks pretty cool :)

PR #309, opened 2024-01-07
Fix imports in Kotlin model generation

MariusVolkhart:
The models project generated Kotlin code that didn't have the correct imports. This happened because the imports were not being tracked as imports in the KotlinCodeGenerator, only as code.


edward3h (approved):
Looks good to me.

If you could add a test case that would catch this, that would be even better.

casid:
Sorry, I just merged a bit too fast. But the change itself is solid!

@edward3h is there a test for the Java way we could adopt for Kotlin?

MariusVolkhart:
I'd probably go the route of writing a test that parses a template, produces generated files, and then compares the generated files against golden versions. Thoughts on that?

The challenge is that the `jte-models` project is the only place where the imports are actually read. The `KotlinCodeGenerator` doesn't care about the list of imports as it generates code directly.

casid:
Indeed, I think writing a test from the jte-models perspective would be ideal, since that's the project that needs the imports as you said.

PR #311, opened 2024-01-07
Use Stream#toList() instead of collect()

MariusVolkhart:
toList was added in Java 16, and some streams can perform a more optimized operation compared to Collectors.toList().


edward3h (approved):
I did not know that `.toList()` could be more efficient than `Collectors.toList()`, but I can believe it. Thanks!

MariusVolkhart:
Comment from Brian Goetz on some details. IIRC, Mark Stuart also had a presentation about it when talking about the Collections API, but I can't find it. :) https://stackoverflow.com/a/65981428/2502247

casid:
I think the main difference is that `collect(Collectors.toList())` currently returns a mutable `ArrayList` (but this could change anytime in future Java versions) and `toList()` returns an immutable list.

This change could potentially break user code for these public methods:

`public List<String> generateAll()`
`public List<String> precompileAll()`
`public List<String> precompile(List<String> names)`

If users attempt to modify the returned lists. Most regular users of the lib don't call these, so it shouldn't be a problem. We should still add a comment to the patch notes though.

PR #313, opened 2024-01-08
Tweak docs

MariusVolkhart:
Fix some Kotlin code, correct an indent level, and provide some more info on how model code generation replaces precompiled templates.

### DIFF
--- a/docs/binary-rendering.md
+++ b/docs/binary-rendering.md
@@ -9,21 +9,25 @@ description: How to use binary rendering for max throughput
 
     Available since jte ^^**1.7.0**^^.
 
-Most template parts are static content, and only a few parts of a template are dynamic. Encoding those static parts repeatedly on every request is wasteful if your web framework sends binary UTF-8 content to the user. jte makes it is possible to encode those static parts at compile time:
+Most template parts are static content, and only a few parts of a template are dynamic. Encoding those static parts repeatedly on every request is wasteful if your web framework sends binary UTF-8 content to the user. jte makes it is possible to encode those static parts when the template is compiled:
 
-=== "Java"
+=== "Java (Application run time)"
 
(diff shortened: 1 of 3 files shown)


casid:
Welcome @MariusVolkhart and thanks for your contributions!

IMHO this note is a bit misleading:
`This replaces the process of [precompiling templates](pre-compiling.md).`

jte-models generates an additional facade so that templates can be called through that facade. But behind the scenes it either calls the templates dynamically and precompiles them on demand (good for development, bad for production), or it calls the previously generated/precompiled template methods (bad for development, good for production).

MariusVolkhart:
Fair point! I was trying to convey that when using the generate stage at build time, you can't also use the pre-compile stage. I found myself spending a fair bit of time trying to wrap my head around the relationship/tradeoff between binary encoding, precompile on-disk, precompile in classpath, and generated models. I'll work on the wording for this.

docs/binary-rendering.md
```diff
@@ -52,7 +52,7 @@ Example usage with `HttpServletResponse`:
 === "Kotlin"
 
     ```kotlin linenums="1"
-    val output = new Utf8ByteOutput()
```
marcospereira, line 55: Good catch!

I'm actually trying to find the time to externalize the code samples in `.java`/`.kt` files that are compiled. It is [supported by MkDocs](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files), and it would be easier to maintain. We can also externalize the template examples.

MariusVolkhart:
@casid I took another stab at this, this time linking more liberally between the strategies (binary encoding, pre-compiled, and facade). I would appreciate you giving this another look and any feedback you might have.

casid:
Looks good, thank you so much!

PR #310, opened 2024-01-07
Reformat the templates for Kotlin model code

MariusVolkhart:
- Change import statements to not have an indent. No functional change.
- Remove unused imports. These were likely copied over from the Java template, but are not needed in Kotlin.
- Dynamic methods, which generate a map, now use Kotlin's buildMap() instead of mapOf(). mapOf() produces extra garbage, as it creates a Pair for each entry, whereas buildMap() modifies the underlying map directly.
- Static methods are now defined as an expression. No functional change.
- In static methods, the template class name is no longer fully qualified, as it's provided as an import. No functional change.
- Remove semicolons - idiomatic Kotlin doesn't use them. No functional change.

PR #316, opened 2024-01-10
ci: Do not cache locally published artifacts

marcospereira:
## What?

These workflows publish artifacts to the local maven repository, which are then cached by GH Actions. Removing them at the end of the workflow prevents them from polluting the cache.

Related to https://github.com/casid/jte/pull/310#issuecomment-1884217056.

It would still be good to delete the existing caches before merging this, and check the effect it would have on #310. Deleting the caches is straightforward if you have GH CLI installed:

```bash
gh cache delete --all
```


casid:
@marcospereira it seems the windows build is still failing?

marcospereira:
> @marcospereira it seems the windows build is still failing?

This is good to go now. :-)

casid:
Thank you @marcospereira!

PR #319, opened 2024-01-16
kte: Fix parameter declaration for generated templates

marcospereira:
## What?

Given the following template:

```html
@param name: String
@param age: Int? = null
```

The generated code for Kotlin's `renderMap` would have:

```kotlin
@JvmStatic fun renderMap(jteOutput:gg.jte.html.HtmlTemplateOutput, jteHtmlInterceptor:gg.jte.html.HtmlInterceptor?, params:Map<String, Any?>) {
	val name = params["name"] as String
	val age = params["age"] as Int?? ?: null
	render(jteOutput, jteHtmlInterceptor, name, age);
}
```

There are two issues here:

1. The double `?` in `Int??` generates a warning (`Redundant '?'`)
2. The section ` ?: null` also generates a warning (`Right operand of elvis operator (?:) is useless if it is null`).

This PR fixes both warnings.

### DIFF
--- a/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngineTest.java
+++ b/jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngineTest.java
@@ -7,6 +7,10 @@ import org.assertj.core.api.AbstractThrowableAssert;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.LocalDateTime;
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


jte-kotlin/src/test/java/gg/jte/kotlin/TemplateEngineTest.java
```diff
@@ -764,6 +768,41 @@ void paramWithDefaultValue() {
         assertThat(output.toString()).isEqualTo("Your age is 10");
     }
 
+    @Test
```
marcospereira, line 771: @casid @edward3h, I'm unsure if this is the best way/place to test it, but let me know what you think.

casid: Looks good, I'd say!

casid:
Thank you @marcospereira!

PR #324, opened 2024-01-26
Fix models extension code generation for tags/templates in subfolders

marcospereira:
## What

When generating code, the jte-models extension does not use the fully-qualified class name for tags/templates in subfolders (therefore, sub-packages). This results in compilation errors.

Also, it adds a new Gradle project that:

- Use jte-models extensions
- Generate Kotlin code

Having a project is better than simply adding more unit tests since it will mimic the experience users have. The new test project is a copy of `test/jte-runtime-cp-test-models-gradle`, but adapted for Kotlin.


casid:
Wow, that was quick. Thank you for fixing this @marcospereira.

marcospereira:
> Wow, that was quick. Thank you for fixing this @marcospereira.

Thanks! I think it would be good to have 3.1.8 release to replace the broken 3.1.7. But it is Friday, so if releasing takes too much of your time, maybe just put a note on the 3.1.7 release about the issue. :-)

casid:
I just released 3.1.8 and left a note on the 3.1.7 release.

PR #330, opened 2024-02-10
fix-escaping-double-quotes

mhdeeb:
Previous version fails`${"\\\"\\"}`. This fix accounts for that.


mhdeeb:
Apologies, I should have added that change to the other PR but this is the first time I PR so I didn't think about it.

casid:
Good catch, nice!

One minor thing. Could you please leave the other test as is and add your test as a second one? I think it would be good to test both cases.

mhdeeb:
Sure, although I think the new test accounts for both if I am not mistaken.

mhdeeb:
Done!

casid:
Thank you!

PR #333, opened 2024-03-01
Update Gradle to 8.6 and add wrapper validation

leonard84:
This PR update Gradle to 8.6 including updated to the generated scripts.
It also adds the `gradle/wrapper-validation-action` action to validate the embedded `gradle-wrapper.jar`.


casid:
Thank you for the upgrade!

PR #336, opened 2024-03-15
Fix link to precompiled templates

mtricht:


casid:
Thanks for the fix @mtricht

PR #339, opened 2024-04-04
maven-plugin: fix kotlin compiler args parameter setting

marcospereira:
Ooops!

There is not much to describe. It was not passing the Kotlin compiler args wrongly as (Java) regular compiler args.


casid:
Aww, good catch, @marcospereira! Thank you for fixing this.

PR #340, opened 2024-04-04
ci: update actions versions to avoid Node 16 warnings

marcospereira:
## What

This is to avoid warnings such as in this build run:

https://github.com/casid/jte/actions/runs/8555571353

I've made minor commits to confirm that each update works as expected.


kelunik (approved):

casid:
Thank you for keeping this up-to-date!

PR #345, opened 2024-04-05
ci: Minor adjustments for dependabot pull requests

marcospereira:
## What?

Two minor changes based on the first PR (#344) opened by dependabot.

- Since it creates a branch in the repository instead of a fork, it runs the workflows twice (one for the PR and one for the push). Therefore, pushes will only run now when it is for the `main` branch.
- It should not try to upload code coverage.

### DIFF
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -9,9 +9,8 @@ on:
       - 'docs/**'
       - 'mkdocs.yml'
   push:
-    paths-ignore:
-      - 'docs/**'
-      - 'mkdocs.yml'
+    branches:
+      - main
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/graalvm.yml
```diff
@@ -9,9 +9,8 @@ on:
       - 'docs/**'
       - 'mkdocs.yml'
   push:
-    paths-ignore:
```
marcospereira, line 12: Since it is the main branch, we always want to run all workflows.

marcospereira:
Updating `codecov/codecov-action` to version v4 broke Codecov uploading because now a token is required:

https://github.com/codecov/codecov-action/blob/v4/README.md#breaking-changes

Example of broken workflow run:

https://github.com/casid/jte/actions/runs/8571119047

> Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token)

The Codecov docs related token setup are here: https://docs.codecov.com/docs/adding-the-codecov-token.

@casid, it seems I can remove [e545b7f](https://github.com/casid/jte/pull/345/commits/e545b7fd9a7c0089aa391ce74add155bee88314e), but you will need to create the Codecov token.

casid:
Hi @marcospereira!

Thanks for the hint with the token, I've added it as a secret.

I'm merging this PR now.

PR #341, opened 2024-04-04
docs: Add Maven Plugin documentation

marcospereira:
## Requirements

This requires #339. I will keep it as a draft until the other PR is merged/closed.

## What

Add a new documentation page to detail the parameters, goals, and usage of the Maven Plugin.

PR #346, opened 2024-04-05
docs: Add new docs page for Gradle Plugin

marcospereira:
Depends on #341.

### DIFF
--- a/docs/gradle-plugin.md
+++ b/docs/gradle-plugin.md
@@ -8,7 +8,7 @@ jte provides a [Gradle Plugin][jte-gradle-plugin] that you can integrate in your
 
 !!! tip "Versions alignment"
 
-    Make sure the jte gradle plugin version matches the jte dependency version. You can create a `jteVersion` in `gradle.properties` to sync the versions easily.
+    Make sure the jte gradle plugin version matches the jte dependency version. You can create a `jteVersion` property in `gradle.properties` to sync the versions easily.
 
 The plugin provides two tasks:
 
@@ -79,7 +79,7 @@ The follow inputs will be picked by the `precompileJte` task:
 
 ## `generateJte` task { #generate-task }
 
-This task generates all template classes in a sources directory. This only generates `.java`/`.kt` files, not `.class` files.
+This task generates all template classes in a sources directory. This only generates `.java`/`.kt` files, but does not compile them to `.class` files.
 
 === "Groovy"
 
--- a/docs/gradle-plugin.md
+++ b/docs/gradle-plugin.md
@@ -1,0 +1,270 @@
+---
+title: Gradle Plugin
+description: How to use the jte Gradle plugin.
+---
+
+jte provides a [Gradle Plugin][jte-gradle-plugin] that you can integrate in your build process to either [precompile templates](pre-compiling.md), or to generate the Java/Kotlin code for your templates.
+
+
+!!! tip "Versions alignment"
+
+    Make sure the jte gradle plugin version matches the jte dependency version. You can create a `jteVersion` in `gradle.properties` to sync the versions easily.
+
+The plugin provides two tasks:
+
+## `precompileJte` task { #precompile-task }
+
+See [Precompiling Templates](pre-compiling.md) for more information. To use and configure this task, you would use:
+
+=== "Groovy"
+
+    ```groovy linenums="1"
+    plugins {
+        id 'java'
+        id 'gg.jte.gradle' version '${jteVersion}'
+    }
+
+    dependencies {
+        implementation('gg.jte:jte:${jteVersion}')
+    }
+
+    jte {
+        precompile()
+    }
+    ```
+
+=== "Kotlin"
+
+    ```kotlin linenums="1"
+    plugins {
+        java
+        id("gg.jte.gradle") version("${jteVersion}")
+    }
+
+    dependencies {
+        implementation("gg.jte:jte:${jteVersion}")
+    }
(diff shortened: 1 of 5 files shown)


marcospereira:
@casid @kelunik, this is now rebased on top of main branch, and ready for review/merge. :-)

docs/gradle-plugin.md
```diff
@@ -0,0 +1,270 @@
...
+| `contentType`           | The content type of all templates. Either `Plain` or `Html`                       | `Html`                                                         |
+| `trimControlStructures` | Trims control structures, resulting in prettier output                            | `false`                                                        |
+| `htmlTags`              | Intercepts the given html tags during template compilation                        | None                                                           |
+| `htmlPolicyClass`       | An [`HtmlPolicy`][html-policy] used to check the parsed HTML at compile time      | None                                                           |
+| `htmlCommentsPreserved` | If HTML comments should be preserved in the output                                | `false`                                                        |
+| `binaryStaticContent`   | If to generate a [binary content](binary-rendering.md) resource for each template | `false`                                                        |
```
kelunik, line 67:
```suggestion
| `htmlCommentsPreserved` | Whether HTML comments should be preserved in the output                                | `false`                                                        |
| `binaryStaticContent`   | Whether to generate a [binary content](binary-rendering.md) resource for each template | `false`                                                        |
```

kelunik: Not sure about these, but `if to` sounds really strange.

docs/gradle-plugin.md
```diff
@@ -0,0 +1,270 @@
...
+| `kotlinCompileArgs`     | Sets additional compiler arguments for `kte` templates                            | None                                                           |
+| `packageName`           | The package name, where template classes are generated to                         | [`Constants.PACKAGE_NAME_PRECOMPILED`][constants-package-name] |
+
+!!! info "About `htmlPolicyClass`"
+
+    The `htmlPolicyClass` will default to `gg.jte.html.OwaspHtmlPolicy` if the content type is `Html`.
+
+!!! warning "Clean all before precompiling"
+
+    The `precompileJte` task cleans the directory containing the compiled template classes every time it runs. In your local development environment, it may make more sense to use [hot reloading](hot-reloading.md).
```
kelunik, line 78: Same comment here as in the other PR: The hint to hot reloading seems unrelated to the `Clean all before precompiling` warning.

docs/gradle-plugin.md
```diff
@@ -0,0 +1,270 @@
...
+
+## `generateJte` task { #generate-task }
+
+This task generates all template classes in a sources directory. This only generates `.java`/`.kt` files, not `.class` files.
```
kelunik, line 82:
```suggestion
This task generates all template classes in a sources directory. This only generates `.java`/`.kt` files, but does not compile them to `.class` files.
```

docs/gradle-plugin.md
```diff
@@ -0,0 +1,270 @@
...
+| `targetDirectory`         | Destination directory to store generated templates.                               | `${project.build.directory}/generated-sources/jte`             |
+| `contentType`             | The content type of all templates. Either `Plain` or `Html`                       | `Html`                                                         |
+| `trimControlStructures`   | Trims control structures, resulting in prettier output                            | `false`                                                        |
+| `htmlTags`                | Intercepts the given html tags during template compilation                        | None                                                           |
+| `htmlCommentsPreserved`   | If HTML comments should be preserved in the output                                | `false`                                                        |
+| `binaryStaticContent`     | If to generate a [binary content](binary-rendering.md) resource for each template | `false`                                                        |
```
kelunik, line 130: Same regarding `if` vs. `whether` applies here.

docs/maven-plugin.md
```diff
@@ -188,7 +188,11 @@ The following parameters are available for this extension:
 
 #### `gg.jte.nativeimage.NativeResourcesExtension` { #native-resources-extension }
 
-See details about it in the [native-image documentation](pre-compiling.md#graalvm-native-image-support). This extension requires no parameters.
+!!! info "Version note"
+
+    Available since jte ^^**1.10.0**^^.
```
kelunik, line 193: What does `^^` do?

docs/gradle-plugin.md
marcospereira: English is not my first language, so it may indeed be strange. However, it is consistent with the javadoc comments. If we decide to adjust here, I would rather adjust the javadoc comments and the Maven plugin docs page.

docs/maven-plugin.md
marcospereira: Underscore. It will be rendered as this:

Screenshot 2024-04-09 at 4 23 09 PM

If I remember correctly, I did it this way when migrating the docs to MkDocs.

docs/gradle-plugin.md
kelunik: I'm not a native speaker either.

docs/maven-plugin.md
marcospereira: To be precise, it is an `insert` that is styled using an underline:

https://facelessuser.github.io/pymdown-extensions/extensions/caret/?h=ins#insert

kelunik: Thanks!

docs/gradle-plugin.md
marcospereira: Grammarly seems okay with this phrasing. I think we can leave it like this.

PR #356, opened 2024-05-07
Add defer and inert to the list of boolean attributes

marcospereira:
I noticed these two were missing, and they are now part of the list at https://meiert.com/en/blog/boolean-attributes-of-html/. The list removed a few attributes, but I kept them for the sake of compatibility.

Also, a small refactor to use `Set.of`. :-)


casid:
Thank you for keeping this up-to-date @marcospereira!

PR #357, opened 2024-05-07
Use Java modern features/APIs

marcospereira:
## What?

This runs [OpenRewrite Java 17 migration recipe](https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-to-java-17) to adopt new/modern Java language features and APIs.

A few of the more extensive changes involve:

- Text blocks: https://openjdk.org/jeps/378
- Pattern matching: https://openjdk.org/jeps/394

## How?

I've added the plugin described in OpenRewrite docs but executed it only for projects that include `jte-parent` as a parent.

### DIFF
--- a/jte-jsp-converter/src/main/java/example/JteContext.java
+++ b/jte-jsp-converter/src/main/java/example/JteContext.java
@@ -17,14 +17,14 @@ public class JteContext {
     public static boolean isEmpty(Object obj) {
         if (obj == null) {
             return true;
-        } else if (obj instanceof String) {
-            return ((String) obj).length() == 0;
-        } else if (obj instanceof Object[]) {
-            return ((Object[]) obj).length == 0;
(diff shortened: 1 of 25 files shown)


casid:
@marcospereira this looks good to me, except that I would like to revert the record commit.

It doesn't bring much benefit compared to the dumb data classes that are already there (except adding getters that are not really needed).

Especially stuff like `public record DirectoryCodeResolver` is really bad. This is not a data class, but public API. The change also breaks the public API for every user who is using `getRoot()` on that class for no reason.

marcospereira:
> It doesn't bring much benefit compared to the dumb data classes that are already there (except adding getters that are not really needed).

Yeah, I was also in doubt about it, mainly because of the compatibility issue, and that is why the PR was a draft. 

It is now gone. :-)

Issue #378, opened 2024-09-22
Spring MVC model not forwarded


blaluc:
It seems that while forwarding requests, the model instance is not injected from the request scope.

e.g.:
```
@Controller
@RequestMapping("/second")
public class OtherController {

	@GetMapping
	public String view2(Model model) {
		model.addAttribute("key2", "second value");
		return "forward:/";
	}
```
```
@Controller
@RequestMapping("/")
public class TestController {

	@GetMapping
	public String index(Model model) {
		model.addAttribute("key1", "first value");
		return "index";
	}
}
```

index.jte:
```
@param String key1
@param String key2

key1: ${key1} <br>
key2: ${key2}
```

By calling http://localhost:8080/second, the `key2` value should be propagated to the template and displayed, but it is not, and the following page is shown instead:
```
key1: first value
key2:
```

The same example with thymeleaf works as expected giving the result:
```
key1: first value
key2: second value
```

blaluc:
it seems that can be solved by adding 
```
this.setExposeRequestAttributes(true);
```
in the `JteViewResolver` class, i.e.:

```
public class JteViewResolver extends AbstractTemplateViewResolver {
...
    public JteViewResolver(TemplateEngine templateEngine, final String templateSuffix) {
        this.templateEngine = templateEngine;
        this.setSuffix(templateSuffix);
        this.setViewClass(JteView.class);
        this.setContentType(MediaType.TEXT_HTML_VALUE);
        this.setOrder(Ordered.HIGHEST_PRECEDENCE);
        this.setExposeRequestAttributes(true);
    }
...
```

casid:
@blaluc thanks for sharing. Your suggested fix looks good to me.

However I'm not a heavy spring user and don't know if this has any unintended side effects.

@atomfrede, @tschuehly would this change be okay with you?

atomfrede:
At first glance it looks okay. Found similar issues with pebble for example. Not sure why it works for thymeleaf out of the box because it looks like the `exposeRequestAttributes` is not set (at least not explicitly).

atomfrede:
[...]

PR #383, opened 2024-09-28
Spring MVC model not forwarded #378

blaluc:
In order to avoid altering the current behavior, a new configuration parameter has been added:

Example:
```
gg.jte.exposeRequestAttributes=true
```

the default value is `false`

This change has been implemented in both `jte-spring-boot-starter-2` and `jte-spring-boot-starter-3`.


jte-spring-boot-starter-3/src/main/resources/META-INF/spring-configuration-metadata.json
```diff
@@ -30,6 +30,12 @@
       "type": "java.lang.String",
       "description": "You can configure the file suffix of jte templates the compiler resolves",
       "sourceType": "gg.jte.springframework.boot.autoconfigure.JteProperties"
+    },
+    {
+      "name": "gg.jte.exposeRequestAttributes",
```
atomfrede, line 35: just to be aligned with the recent change:

```suggestion
      "name": "gg.jte.expose-request-attributes",
```

PR #408, opened 2024-12-01
Fix Examples on Localization Doc Page

cilhao:
On documentation Localization page the examples shown for both Java and Kotlin are calling the wrong **gg.jte.support.LocalizationSupport.lookup()** method name "localize" and the correct is "lookup".


casid:
Welcome @cilhao!

No, `localize` is the correct method.

Lookup is an internal method that needs to be implemented by the particular localization framework you're using.

In templates, `localize` should be used.

PR #410, opened 2024-12-05
ci: Re-arrange workflows to remove duplication

marcospereira:
## Motivation

There are two primary motivations here:

1. **Allow to add jobs to the workflows easily**. For example, it should be easy to test against a new JDK version or a new Kotlin version. We currently do that via matrices and steps. This brings all to matrices.
1. **Remove duplication**: there are duplications regarding the matrix configuration and steps to set up caching.

The final workflow looks like this:

Screenshot 2024-12-04 at 11 23 50 PM

Reference: https://github.com/marcospereira/jte/actions/runs/12173115754

### Merging Graal and Maven workflows

They are mostly copied and pasted from each other, with minor differences, and bringing them together into a single workflow file helps to later maintain consistency in a single place.

### Jobs dependencies

Previously, all the jobs were executed simultaneously, meaning that even if the code does not pass unit tests, it still runs coverage, graal tests, etc. Now, there is a dependency between jobs, so we only execute more expensive workflow sections if the tests pass.

This hasn't added much to the workflow total time.

### Reusable actions

Two new reusable actions exist to set up Gradle caching and remove locally published artifacts before caching. Although it is possible to add a few other reusable actions for steps that seem repetitive, I've decided to add only these two because they don't add logic to the build. Instead, they are mainly CI infrastructure.

## Future work

I will build on top of this to add support for Kotlin 2.1.0. It is currently broken for the Gradle plugin.


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -18,4 +18,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: gradle/wrapper-validation-action@v3
+      - uses: gradle/actions/wrapper-validation@v4
```
marcospereira, line 21: An small bonus to update this workflow to use the new wrapper-validation action.

.github/workflows/maven.yml
```diff
@@ -94,7 +170,7 @@ jobs:
         if: contains(matrix.os, 'win') == false
         run: chmod +x ./mvnw
       - name: Build with Maven
-        run: ./mvnw verify --file pom.xml -Pcoverage
+        run: ./mvnw verify --file pom.xml -Pcoverage,kotlin-1.9
```
marcospereira, line 173: `kotlin-1.9` needs to be explicitly activated here, as [documented here](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#Explicit_profile_activation):

> This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.

test/jte-runtime-test-kotlin/pom.xml
```diff
@@ -69,6 +68,22 @@
                 </plugins>
             </build>
         </profile>
+        <!-- Kotlin versions -->
```
marcospereira, line 71: I don't like having this profile in two places, but I thought it was better than moving it to the parent pom.

nedtwigg:
This looks great to me!

casid:
Hi @marcospereira, this looks great! I really like, that we now have everything in one build file together.

And sorry for the late reply, I just returned from a long vacation.

PR #411, opened 2024-12-05
WIP: Add support to Kotlin 2.1

marcospereira:
The main work needs to be done on the Gradle plugin.

This is built on top of #410.


jte-kotlin/src/main/java/gg/jte/compiler/kotlin/KotlinClassCompiler.java
```diff
@@ -39,7 +39,7 @@ public void compile(String[] files, List<String> classPath, TemplateConfig confi
         K2JVMCompiler compiler = new K2JVMCompiler();
 
         SimpleKotlinCompilerMessageCollector messageCollector = new SimpleKotlinCompilerMessageCollector(templateByClassName, config.packageName);
-        ExitCode exitCode = compiler.exec(messageCollector, new Services.Builder().build(), compilerArguments);
+        ExitCode exitCode = compiler.exec(messageCollector, Services.EMPTY, compilerArguments);
```
GregoryBevan, line 42: Hi,
Is it possible to make this correction in another PR to avoid the following error :
```
java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.cli.common.ExitCode org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.exec(org.jetbrains.kotlin.cli.common.messages.MessageCollector, org.jetbrains.kotlin.config.Services, org.jetbrains.kotlin.cli.common.arguments.CommonToolArguments)'
```
I got this after migrating my jte project to version 2.1.0 of Kotlin

casid:
Thank you, for looking into this @marcospereira!

I've just merged #410.

If you need any help, let me know!

marcospereira:
Thanks, @casid .

I'm coming back from vacations myself, and will take a look at this later. But so far, I think there is some dependency conflict/issue breaking things for Kotlin 2.1.

casid:
@marcospereira I'm currently trying to reproduce the issues locally but I am a bit lost.

This build breaks already at the maven stage, when building jte-kotlin (https://github.com/casid/jte/actions/runs/12193098684/job/34014872139?pr=411). I checked out the support-kotlin-2.1.x branch but could not reproduce the issue. I hardcoded kotlin version 2.1.0, because for some reason the activation profile gave me class incompatability error (but different to the errors in the build).

Would it help with complexity, if we say Kotlin 2.1 or greater will be required for the next jte release, and whatever Gradle version is compatible with that (I think 8.6)?

checketts:
I'm supportive of making Kotlin 2.1 a requirement if it is needed to move forward.

checketts:
@casid What can I do to help with this PR? Your last comment seems to say the build works for you. Do you need me to build it locally and test it against my projects? Is there a RC candidate I could pull into my project?

casid:
@checketts sorry for the late reply, I currently don't have much time for open source projects.

I just tagged you in this PR, can have a look if this would work for you as well?

https://github.com/casid/jte/pull/431

checketts:
@casid I totally understand. I'm glad to try to lend a hand in this (though I am not particularly experienced with plugins etc). Testing the packages from #431 worked for me.

edward3h:
I've merged some more recent PRs relating to Kotlin changes. Is this one still needed?

PR #417, opened 2024-12-22
Use spring-configuration-metadata.json generated by spring boot configuration processor instead of manually creating it

genuss:
Hello!

I've just come across your project and so far I love it! While browsing through the code I found that you don't use spring-boot-configuration-processor to generate configuration properties netdata. Is it intentional? If not, I'd like to propose a small contribution. Instead of writing these files manually, it's better to have them generated from javadocs. This have some advantages:
* less code overalll.
* less error-prone. For example, the file for spring boot 2 states there is a `gg.jte.development-mode` property. This is not the case though, it's only supported in spring boot 3.
* Users of intellij idea community edition can see documentation by reading javadocs. As far as I understand, reading of ...metadata.json` files is an ultimate-only feature.

What do you think about it?


casid:
Welcome @genuss and thanks for the heads up!

I'm not very deep into Spring, so can't really give you meaningful feedback.

@atomfrede what's your opinion on this?

atomfrede:
To make my 👍 more explicit, let's do it, looks great, was not aware of the annotation processor

PR #431, opened 2025-02-27
Only support Kotlin 2.1

TheKinrar:
As an alternative to https://github.com/casid/jte/pull/411, this PR removes support for Kotlin < 2.1 and simply requires Kotlin 2.1 going forward. I guess this might not be what you want, but works for me so I'm just pushing the (very minor) changes here in case it's of interest to someone!

I also have packages from this PR available on my repo: https://github.com/TheKinrar/jte


casid:
Thanks @TheKinrar for the PR. Awesome, that makes everything a bit easier.

@marcospereira, @checketts what do you think, shall we move forward with this, so that we can have a release that works with the latest and greatest Kotlin again?

checketts:
Thanks for the packages @TheKinrar . I was able to pull them down and verify that my project is working with Kotlin 2.1.0 and JTE.

So @casid  yes, please move forward with this!

TheKinrar:
@casid Gradle decided to change the line endings of the gradlew.bat files when I updated it, thus the now significant diff although I did not change much. I can restore the old gradlew files if needed. Old Gradle versions caused a conflict because of their old embedded Kotlin version.

casid:
Looks good, I'm merging this now!

PR #432, opened 2025-03-03
Java compiler error matching

SebastianSchlecht:
When compiling a template containing a mix of warnings and errors the first warning/error will be matched by `getErrorMessage` which might not be the desired error message.
This leads to a misleading TemplateException like

```
10:32:17,860 ERROR Failed to execute goal gg.jte:jte-maven-plugin:3.1.13:precompile (web-templates) on project example: Execution web-templates of goal gg.jte:jte-maven-plugin:3.1.13:precompile failed: Failed to compile template, error at some/file/with/warning.jte:35
```
with the culprit buried deeper inside the log.

This proposed change will only match messages of the type 'error' and therefore shows the message responsible for the failed compilation on top.


casid:
Awesome, thanks @SebastianSchlecht!

Issue #412, opened 2024-12-06
JTE plugin failing for me with Kotlin 2.1.0 upgrade


Jerbell:
So...

I have multi-module project. When I try to build my non-JTE project with the JTE plugin enabled I get the warning as defined in #407  but I also get the following error:
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not initialize class org.jetbrains.kotlin.gradle.report.ConfigureReporingKt
> Exception java.lang.IncompatibleClassChangeError: class org.jetbrains.kotlin.build.report.metrics.GradleBuildTime can not implement org.jetbrains.kotlin.build.report.metrics.BuildTime, because it is not an interface (org.jetbrains.kotlin.build.report.metrics.BuildTime is in unnamed module of loader org.g [in thread "Daemon worker"]
```

If I disable the JTE plugin then my non-JTE module builds successfully. This seems to suggest there is a problem with the plugin.

Thanks very much & let me know if I can be of assistance.

marcospereira:
Can you share the full output when you ran the build with `--info`?

I want to confirm that you may be facing the same issue I observed. By the way, I'm working on #411. It may take a little while since it is the end of the year already.

marcospereira:
Also, maybe it is better to close this issue and add comments to #407.

Jerbell:
> Also, maybe it is better to close this issue and add comments to #407.

Maybe. #407 is a warning anyone though I guess a fix will resolve both.
I'll give you more info.

Jerbell:
<details><summary>Info</summary>Initialized native services in: C:\Users\Julian\.gradle\native
Initialized jansi services in: C:\Users\Julian\.gradle\native
To honour the JVM settings for this build a single-use Daemon process will be forked. For more on this, please refer to https://docs.gradle.org/8.11.1/userguide/gradle_daemon.html#sec:disabling_the_daemon in the Gradle documentation.
[...]

Issue #407, opened 2024-11-30
Kotlin 2.1 warning due to JTE accessing hidden compiler symbols.


sureshg:
Kotlin 2.1 now throws a warning and will soon prohibit accessing `kotlin-compiler-embeddable` from build scripts. I am now starting to get warning messages from build scripts which use the`JTE` plugin. When i checked the dependencies for the build plugin, `kotlin-compiler-embeddable` is used by JTE and the Kotlin 2.1 doc suggests the following [workaround](https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api) to fix the issue.

```
org.jetbrains.kotlin:kotlin-compiler-embeddable:1.9.10 -> 2.1.0
\--- gg.jte:jte-kotlin:3.1.15
     \--- gg.jte:jte-gradle-plugin:3.1.15
          \--- runtimeClasspath
```

```
> Configure project :shared
w: The artifact `org.jetbrains.kotlin:kotlin-compiler-embeddable` is present in the build classpath along Kotlin Gradle plugin.
This may lead to unpredictable and inconsistent behavior.
For more details, see: https://kotl.in/gradle/internal-compiler-symbols

> Configure project :web
w: The artifact `org.jetbrains.kotlin:kotlin-compiler-embeddable` is present in the build classpath along Kotlin Gradle plugin.
This may lead to unpredictable and inconsistent behavior.
For more details, see: https://kotl.in/gradle/internal-compiler-symbols
```

casid:
Is this fixed in jte 3.2.0?

sureshg:
@casid thanks for 3.2.0 release, but i can still see these warnings. 

Instead of direclty pulling [this](https://github.com/casid/jte/blob/main/jte-kotlin/pom.xml#L22-L26), should we we need to use something like https://kotlinlang.org/docs/whatsnew21.html#using-the-gradle-workers-api

```
org.jetbrains.kotlin:kotlin-compiler-embeddable:2.1.20
\--- org.jetbrains.kotlin:kotlin-bom:2.1.20
     +--- runtimeClasspath
     \--- com.github.ben-manes:gradle-versions-plugin:0.52.0 (requested org.jetbrains.kotlin:kotlin-bom:2.0.21)
          \--- com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.52.0
               \--- runtimeClasspath

[...]

PR #448, opened 2025-04-10
Gradle Plugin - isolated Kotlin compiler

Nickoriuk:
Hi, thank you for the great project!

I was having some compatibility issues with JTE 3.2.0 with the Kotlin version used in my projects. It appears to be related to the Kotlin 2.1 changes (https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api) that have been discussed in #412, #407.

This change should allow the Kotlin compiler for JTE to be independent of the version used for downstream projects. With my projects using JTE, I can now run run this branch's artifacts with Kotlin 2.1.20, where before I could only use Kotlin 2.1.10 with JTE 3.2.0. The Kotlin gradle warning:

> w: The artifact `org.jetbrains.kotlin:kotlin-compiler-embeddable` is present in the build classpath along Kotlin Gradle plugin.
This may lead to unpredictable and inconsistent behavior.
Please remove the `org.jetbrains.kotlin:kotlin-compiler-embeddable` artifact from the build classpath.
See https://kotl.in/gradle/internal-compiler-symbols for more details.

...in my downstream project is now also gone.

This PR has 3 commits:

- the first adds a copy of `test/jte-runtime-cp-test-models-gradle-kotlin` to `test/jte-runtime-cp-test-models-gradle-kotlin-versionmismatch`, with the only other change being build.gradle -> build.gradle.kts. This test project builds fine.
- The second changes the Kotlin plugin version to 2.1.20 (JTE still using 2.1.10) for the new test directory, causing the test project to fail.
- The third commit fixes the tests by updating the plugin and dependency configuration.

I'm not a Gradle plugin expert so hopefully these changes make sense! I'm happy to make any adjustments.


Nickoriuk:
This diff is based on top of #445 - ran a quick check on my local and confirmed that reverting it causes this PR to break - for example `test/kte-runtime-test-gradle-kotlin-convention` would give the following error:

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':precompileJte'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:130)
...<snip>...
Caused by: java.lang.NoClassDefFoundError: org/jetbrains/kotlin/gradle/internal/analyzer/CompilationErrorException
	at gg.jte.compiler.kotlin.KotlinClassCompiler.compile(KotlinClassCompiler.java:39)
```

---

These changes are related to #445 in that both changesets are intended to solve the issue described in the [Kotlin 2.1 release notes](https://kotlinlang.org/docs/whatsnew21.html#compiler-symbols-hidden-from-the-kotlin-gradle-plugin-api) - this just addresses a few changes from that page that were not implemented in the PR (from what I can tell).

In that page, there are four code blocks showing the steps to migrate to the Gradle Workers API:

1. Ensuring the compiler is in the `compileOnly` scope
2. Using `WorkAction`
3. Using `WorkerExecutor` with classloader isolation
4. Configuring the compiler classpath in `Plugin#apply`

From what I can tell, #445 implements 2) and 3). I see some other changes in there too but it doesn't match the docs above - not sure whether those are related specifically to KGP isolation. I don't see implementation of 1) and 4).

This PR implements 1) by excluding the transitive `kotlin-compiler-embeddable` dependency from `jte-kotlin` from the runtime classpath, re-adding it in compileOnly scope, and 4) with the `jteKotlinCompiler` configuration.

---

I believe the tests passed with just 2) and 3) because the Kotlin Gradle Plugin and JTE Gradle Plugin in all tests both pulled in `org.jetbrains.kotlin:kotlin-compiler-embeddable:2.1.10` so the compilers on the classpath were binary compatible, but since 1) and 4) were not done, a different plugin version can still cause issues, indicating compiler isolation is not fully working. This and #445 together should address the issue more fully.

schowave:
Thanks for pointing it out and for the implementations and the Tests. I Really did not Set the dependency to compileOnly…

edward3h (approved):

sureshg:
@edward3h apologies for the direct ping. Are there any plans for a JTE release soon?

edward3h:
@casid can you make a release?

casid:
This is now live with jte 3.2.1

Thank you everyone to make this happen. That's pretty big for Kotlin / Gradle users!

schowave:
Thank you for the release, our build is now free from those warnings :-)

Issue #370, opened 2024-08-06
JteModels extension with a 'Plain' (non-HTML engine)


checketts:
In following the JteModels documentation I seem to be able to only create Html based templates.

However, I need to create a few text (plain) templates.

Does that support exist? Any hint on how that might be configured?

edward3h:
Hi @checketts,
I'm the author of jte-models. I'm kinda busy but I will try to get back to you at the weekend.

checketts:
Thanks for the `jte-models`!

Here is my imagined way of making it work:

```kotlin

    //Generate HTML based templates
    jteExtension("gg.jte.models.generator.ModelExtension") {
        property("language", "Kotlin")
        property("excludePattern", ".*_text.*")
    }

    //Generate Text based templates
    jteExtension("gg.jte.models.generator.ModelExtension") {
        property("language", "Kotlin")
        property("package", "gg.jte.generated.precompiled.text") // Will place the `Templates` classes in this package
        property("contentType", "plain")
        property("includePattern", ".*_text.*")
    }
```

Note the new `package` and `contentType` options.

edward3h:
jte-models can generate models for plain text templates. It uses the `contentType`  specified on the `jte { }` block in Gradle.

This does make it difficult to use both HTML and plain templates at the same time. I think you would have to put them in separate submodules.

edward3h:
Sorry for the delay in responding, by the way.

edward3h:
It would not be too hard to make your imagined configuration work, but it would require some code changes.

checketts:
Thanks for the response. My usecase is generating emails, since email has an HTML envelope and a plaintext. So separate modules wouldn't necessarily work for me.

Are you aware of a way to pass separate `jteExtension` config to a separate Gradle task? Perhaps I could investigate that route.

edward3h:
The JTE Gradle plugin currently makes some assumptions that will prevent you configuring a separate task with different settings. It would be good to fix, but I don't have time available to work on that at the moment.

sysmat:
doc for jte-models https://jte.gg/jte-models/  is un clear, and not working with maven, no StaticTemplates is generated

sysmat:
- when switching to `generate-sources` maven compile error `cannot find symbol`
- package gg.jte.models.runtime does not exist even if I add dependency jte-runtime

edward3h:
[...]

PR #457, opened 2025-05-18
Multiple gradle builds

edward3h:
Addresses #370 by supporting multiple builds from the gradle plugin.


edward3h:
TODO

* Test case using jte-models, to make sure it really addresses #370 
* Documentation updates

edward3h:
I stalled working on this. I agree with comments elsewhere that it should support multiple Gradle tasks, rather than a single task doing multiple builds.

PR #474, opened 2025-08-23
Update maven-surefire-plugin

asvanberg:
2.22.2 is over 6 years old.


edward3h:
All tests pass - seems good enough to me.

Issue #490, opened 2025-11-04
JTE spring boot starter conflicts with other plugins adding view resolvers


raoulvdberge:
Hello there,

When the JTE spring boot integration is used, the view resolvers of the [HTMX spring boot integration](https://github.com/wimdeblauwe/htmx-spring-boot) break (meaning, the views never get precedence when using something like `redirect:htmx:location:...`).

After investigation, it seems like this happens because JTE registers themselves with the highest priority, which is not something that other extensions do too (thymeleaf or freemarker for example).

For more context and detailed analysis, please see: https://github.com/wimdeblauwe/htmx-spring-boot/issues/180

Problematic code: https://github.com/casid/jte/blob/main/jte-spring-boot-starter-2/src/main/java/gg/jte/springframework/boot/autoconfigure/JteViewResolver.java#L19

PR #491, opened 2025-11-04
#490 do not set order in JteViewResolver

casid:


atomfrede:
Looks fine and makes totally sense to not Set the order

PR #500, opened 2026-01-08
Improve the pre-compiling docs

eddiclin:
Since we use precompiled binary content parts, we need to copy those resources next to the compiled class files (the maven plugin doesn't do that for us).
Reference URL: https://github.com/casid/jte-spring-boot-demo/blob/master/pom.xml#L72


edward3h:
Thank you for making a contribution 😄

PR #502, opened 2026-01-13
Make generated Kotlin classes internal

MariusVolkhart:
There are two reasons for this. The first is that there's no reason for these to be public. They are consumed entirely in the current compilation unit, and there's no reason for external code to ever reference these. This comes up when working in a large Spring/Micronaut project where smaller modules act as libraries to a final aggregate.

The other reason to do this is that by making this classes public, it prevents them from being used with data classes that the developer declared as internal. The compiler will complain that the generated code exposes an internal class in a public API.


MariusVolkhart:
I should note that this also allows jte to be used with Kotlin projects that have `explicitApi()` enabled

Issue #508, opened 2026-01-23
Spring Boot view resolver priority issue


edward3h:
Workaround shared by @schowave , which also describes the problem.

> I also had the problem that the templates were not rendered, this was my solution:
> 
> ```java
> import gg.jte.springframework.boot.autoconfigure.JteViewResolver
> import org.springframework.beans.factory.annotation.Autowired
> import org.springframework.context.annotation.Configuration
> import org.springframework.core.Ordered
> 
> /**
>  * Configuration for JTE view resolver priority.
>  *
>  * In JTE 3.2.2, the explicit order setting was removed from JteViewResolver to fix conflicts
>  * with HTMX (PR #490/#491). However, this causes InternalResourceView to take precedence
>  * in Spring Boot 4.0, preventing JTE templates from rendering.
>  *
>  * This configuration explicitly sets the JTE view resolver's order to be higher priority
>  * than InternalResourceViewResolver, ensuring JTE templates are resolved first.
>  */
> @Configuration
> class JteViewResolverConfiguration {
>     @Autowired
>     fun configureJteViewResolver(jteViewResolver: JteViewResolver) {
>         // Set JTE resolver to highest priority to ensure it resolves before InternalResourceView
>         jteViewResolver.order = Ordered.HIGHEST_PRECEDENCE
>     }
> }
> 
> ```

 _Originally posted by @schowave in [#505](https://github.com/casid/jte/issues/505#issuecomment-3789245782)_

casid:
Hmm, I wonder how other template engine Spring Boot integrations solve this.

Ideally, this should be something taken care of by the spring boot starter.

Gadnex:
I had a look around and found this code for Thymeleaf in spring boot:
https://github.com/spring-projects/spring-boot/blob/v4.0.2/module/spring-boot-thymeleaf/src/main/java/org/springframework/boot/thymeleaf/autoconfigure/ThymeleafAutoConfiguration.java

```java
@Bean
@ConditionalOnMissingBean(name = "thymeleafViewResolver")
ThymeleafViewResolver thymeleafViewResolver(ThymeleafProperties properties,
		SpringTemplateEngine templateEngine) {
	ThymeleafViewResolver resolver = new ThymeleafViewResolver();
	resolver.setTemplateEngine(templateEngine);
	resolver.setCharacterEncoding(properties.getEncoding().name());
	resolver.setContentType(
			appendCharset(properties.getServlet().getContentType(), resolver.getCharacterEncoding()));
[...]

PR #517, opened 2026-02-05
Add Spring Boot 4 Starter modules

### LINKED ISSUE #512: Add new module jte-spring-boot-starter-4
It would probably be a nice-to-have for a dedicated module targeting [Spring Boot](https://spring.io/projects/spring-boot) `4` (`4.0.2`)

- Add the module
- Add documentation

The `jte-spring-boot-starter-3` works okay with `Boot 4`, so it would most likely just be a convenience to pull in the right dependencies automatically.

### LINKED ISSUE #514: Skip ViewResolver registration based on property
I'm using jte for templating emails and reports. But my app is all REST API. It would be nice if it's possible to skip registering the bean JteViewResolver based on a property: `gg.jte.web=false` that should disable any MVC related components.

Ingramz:
Fixes #512
Fixes #514

Adds two new packages `jte-spring-boot-starter-4` and `jte-spring-boot-starter-4-core` with the following dependency hierarchy.
```
+-- jte-spring-boot-starter-4
  `-- jte-spring-boot-starter-4-core
    `-- jte
```

`jte-spring-boot-starter-4` is intended to be the main starter package providing auto-configuration for everything, similarly to `jte-spring-boot-starter-3`. It only hosts the auto-configuration for `ViewResolver`s (servlet & reactive) and pulls in the auto-configuration for `TemplateEngine` from `jte-spring-boot-starter-4-core` dependency.

`jte-spring-boot-starter-4-core` is a new intermediary package which includes auto-configuration for just the `TemplateEngine`. A developer can use this package directly if they do not wish to auto-configure the `ViewResolver` for their app. This can be useful for applications where JTE is not used for MVC views. For example email and document templates.

Alternatively the structure could also split the servlet and reactive `ViewResolver`s to their respective submodules and the main module would be an empty parent that was simply a metapackage for pulling in everything. I do not see the need to do this at least for now, but theoretically it could be done in future without introducing a breaking change.

The `-core` suffix was simply the first one that came to my mind, I have no preference what it should be (`-base`, `-template`, `-your-alternative-here`).

`JteProperties` is part of the `-core` package and contains also properties that are only applicable to `ViewResolver`s. This appears to be a limitation of Spring Boot where multiple `@ConfigurationProperties` beans with same prefix (`gg.jte`) aren't allowed exist, so I was unable to break these away from the `-core` package. Moving some of the properties to a different prefix would be a breaking change that I am currently trying to avoid. The only drawback is that `-core` users would also see properties in their IDE that do not apply to them.

Other changes from `jte-spring-boot-starter-3` in no particular order:

* Moved servlet `ViewResolver` from `JteAutoConfiguration` to dedicated `ServletJteAutoConfiguration` (mirroring `ReactiveJteAutoConfiguration`). Other classes (`JteView` and `JteViewResolver`) were not renamed in case they are part of someones user code.
* Added `@NonNull` annotations to methods with `@Override` where they were missing.
* Migrated `@NonNull` annotations to JSpecify (Spring Boot 4 change).
* Fixed `DataBufferFactory.allocateBuffer()` deprecation by specifying default buffer size.
* Removed explicit `jakarta.servlet-api` dependency version (`spring-boot-dependencies` provides the applicable version automatically).
* Added `@ConditionalOnWebApplication(type = Type.SERVLET)` to `ServletJteAutoConfiguration`.
* Added `@ConditionalOnWebApplication(type = Type.REACTIVE)` to `ReactiveJteAutoConfiguration`.
* Changed star imports to single class imports (different files used different styles, but majority seemed to be star imports, so I'm willing to change all of them back if desired).
* Copied docs and added section about the `-core` package.
* Removed explicit `gg.jte:jte` dependency from install instructions (to align with https://start.spring.io)
* Removed duplicated getters and setters for `usePrecompiledTemplates` in `JteProperties`
* Updated `maven-compiler-plugin` to latest so the correct dependency version for `spring-boot-configuration-processor` is automatically requested in build plugins section
* Changed `spring-boot-configuration-processor` from dependency to [build plugin configuration](https://docs.spring.io/spring-boot/specification/configuration-metadata/annotation-processor.html#appendix.configuration-metadata.annotation-processor.configuring)
* Removed direct dependency on `spring-boot-autoconfigure` as it is always transitively resolved from `spring-boot-starter`
* Added auto-configuration tests for TemplatEngine and ViewResolvers
* Added sample application tests to catch cases such as #508

Feel free to course-correct as necessary.

Some of these changes could be also made to `jte-spring-boot-starter-3`, but for now I limited my changes only to adding the new packages.

I did test the newly created packages locally and they seem to have the desired behavior, but it would be appreciated if somebody else could also give a try.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,8 @@ Temporary Items
 /jte/jte/
 /jte/jte-classes/
 /jte-kotlin/jte-classes/
+/jte-spring-boot-starter-4/jte-classes/
+/jte-spring-boot-starter-4-core/jte-classes/
 /jte-watcher/jte-classes/
 /test/*/jte-classes/
 .run
--- a/docs/spring-boot-starter-4.md
+++ b/docs/spring-boot-starter-4.md
@@ -1,0 +1,106 @@
+---
+title: jte Spring Boot 4 Starter
+description: A Spring Boot 4 Starter to swiftly integrate jte
+---
+
+# Spring Boot Starter
+
+This [starter](https://docs.spring.io/spring-boot/4.0/reference/using/build-systems.html#using.build-systems.starters) is compatible with Spring Boot 4.x!
+
+## Add required dependencies
+
+=== "Maven"
+
+    ```xml linenums="1"
+    <dependency>
+        <groupId>gg.jte</groupId>
+        <artifactId>jte-spring-boot-starter-4</artifactId>
+        <version>{{ latest-git-tag }}</version>
+    </dependency>
+    ```
+
+=== "Gradle"
+
+    ```groovy linenums="1"
+    implementation "gg.jte:jte-spring-boot-starter-4:{{ latest-git-tag }}"
+    ```
+
+## Usage
+
+The starter configures a `org.springframework.web.servlet.ViewResolver` and a jte Template engine. Now you can return a string, pointing to template file name and the resolver will take care to instantiate the view and render the template.
+
+!!! info
+
+    By default, the templates are expected at `src/main/jte`.
+
+```html linenums="1"
+@import com.example.demo.DemoModel
+
+@param DemoModel model
+
+Hello ${model.text}!
+```
+
+```java linenums="1"
+@GetMapping("/") 
+public String view(Model model, HttpServletResponse response) {
+    model.addAttribute("model", new DemoModel("Hello World"));
+    return "demo";
+}
+```
+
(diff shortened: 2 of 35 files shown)


edward3h:
Could you add an automated test? It doesn't need to be exhaustive - just make sure a template can get rendered through this framework. This will mean if a bug comes up later, it is easy to add another test case for it.

atomfrede:
Nice work!

Issue #526, opened 2026-03-11
Cryptic error message when rendering a precompiled HTML template as ContentType.Plain


iTob191:
When rendering a template with `ContentType.Plain` that was pre-compiled for `ContentType.HTML`, I'm getting a pretty cryptic error message:
```
gg.jte.TemplateException: Failed to render ...
	at gg.jte.TemplateEngine.handleRenderException(TemplateEngine.java:259) ~[jte-runtime-3.2.2.jar:na]
	at gg.jte.TemplateEngine.render(TemplateEngine.java:232) ~[jte-runtime-3.2.2.jar:na]
...
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:108) ~[na:na]
	at java.base/java.lang.reflect.Method.invoke(Method.java:565) ~[na:na]
	at gg.jte.runtime.Template.renderMap(Template.java:58) ~[jte-runtime-3.2.2.jar:na]
	at gg.jte.TemplateEngine.render(TemplateEngine.java:230) ~[jte-runtime-3.2.2.jar:na]
```

A better error message for this problem was implemented in https://github.com/casid/jte/commit/bafa85de3edfa82397a0d73e44812b34e3067ecb, but only when using `render(String, Object, TemplateOutput)` and not when using `render(String, Map<String, Object>, TemplateOutput)`.

kelunik:
Thanks for reporting!

Do you want to send a pull request?

iTob191:
See #529

iTob191:
Resolved by #529

PR #529, opened 2026-03-16
#526 Provide better exception message if template engine is initialized with wrong content type

iTob191:
Follow-up to bafa85de3edfa82397a0d73e44812b34e3067ecb

### DIFF
--- a/jte-runtime/src/main/java/gg/jte/runtime/Template.java
+++ b/jte-runtime/src/main/java/gg/jte/runtime/Template.java
@@ -37,19 +37,17 @@ public final class Template {
         } catch (InvocationTargetException e) {
             throw e.getCause();
         } catch (IllegalArgumentException e) {
-            if (render.getParameterTypes()[0] == HtmlTemplateOutput.class && !(output instanceof HtmlTemplateOutput)) {
-                throw new TemplateException("The template " + name + " was compiled with ContentType.Html, but the template engine was initialized with ContentType.Plain. Please initialize the template engine with ContentType.Html.", e);
-            } else {
-                String expectedType = render.getParameterTypes()[2].getName();
-                String actualType = param != null ? param.getClass().getName() : null;
-                String message = "Failed to render " + name + ", type mismatch for parameter: Expected " + expectedType + ", got " + actualType;
+            checkWrongContentType(output, e);
 
-                if (isMangledKotlinRenderMethod(render)) {
-                    message += "\nIt looks like you're rendering a template with a Kotlin value class parameter. To make this work, pass the value class parameter in a map.\nExample: templateEngine.render(\"" + name + "\", mapOf(\"myValue\" to MyValueClass(), output)";
-                }
+            String expectedType = render.getParameterTypes()[2].getName();
+            String actualType = param != null ? param.getClass().getName() : null;
+            String message = "Failed to render " + name + ", type mismatch for parameter: Expected " + expectedType + ", got " + actualType;
 
-                throw new TemplateException(message, e);
+            if (isMangledKotlinRenderMethod(render)) {
+                message += "\nIt looks like you're rendering a template with a Kotlin value class parameter. To make this work, pass the value class parameter in a map.\nExample: templateEngine.render(\"" + name + "\", mapOf(\"myValue\" to MyValueClass(), output)";
             }
+
+            throw new TemplateException(message, e);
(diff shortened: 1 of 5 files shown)


edward3h (approved):

jte-runtime/src/main/java/gg/jte/runtime/Template.java
```diff
@@ -58,6 +56,14 @@ public void renderMap(TemplateOutput output, HtmlInterceptor htmlInterceptor, Ma
             renderMap.invoke(null, output, htmlInterceptor, params);
         } catch (InvocationTargetException e) {
             throw e.getCause();
+        } catch (IllegalArgumentException e) {
+            checkWrongContentType(output, e);
```
casid, line 60: We must throw the exception here, in case it is not the wrong content type exception.

I think `throw e;` after `checkWrongContentType`

Would be good to have a test for it too, I guess.

test/jte-runtime-test/src/test/java/gg/jte/WrongContentTypeTest.java
```diff
@@ -33,19 +33,21 @@ void setUp() {
...
+    @Test
+    void wrongContentTypeInRenderMap() {
+        renderingFailsBecauseOfWrongContentType(() -> templateEngine.render("helloWorld.jte", Map.of("model", model), output));
     }
 
     @SuppressWarnings("SameParameterValue")
-    private AbstractThrowableAssert<?, ? extends Throwable> thenRenderingFailsWithException(String templateName) {
-        Throwable throwable = catchThrowable(() -> whenTemplateIsRendered(templateName));
-        return assertThat(throwable).isInstanceOf(TemplateException.class);
+    private void renderingFailsBecauseOfWrongContentType(Runnable whenTemplateIsRendered) {
```
casid: Please follow the naming conventions and name this method **then**RenderingFailsWith...

casid (changes requested):
Thanks for the PR, can please check my comments?

jte-runtime/src/main/java/gg/jte/runtime/Template.java
iTob191: Good catch. I've added a test for it.

~~While doing so, I've noticed that the handling of `IllegalArgumentException` in `render(..)` assumes that every exception that is not caused by a wrong content type, is caused by a wrong template parameter and generates a respective `TemplateException`. However, this assumption is wrong if the user code throws an `IllegalArgumentException` and would probably lead to a weird exception (e.g. `Failed to render template.jte, type mismatch for parameter: Expected String, got String`).~~

iTob191: While testing a little further, I've noticed that the [additional test I've added](https://github.com/casid/jte/pull/529/changes/c32c626890473c338950186a6d40513876840928..3eb7b8ec290b27f67e73489d255096167534702d) does not actually test the added [`throw e;`](https://github.com/casid/jte/pull/529/changes#diff-92cf9e42f2e231accb42aa4d86207e79031f1936d29e9ad12a62e07cbffc4153R61) because the exception is wrapped in an `InvocationTargetException`. Should I remove the test again?

Also, I'm not really sure how to test the `throw e;` because I can't think of any input that would trigger it.

test/jte-runtime-test/src/test/java/gg/jte/WrongContentTypeTest.java
```diff
@@ -32,6 +33,11 @@ void setUp() {
         model.hello = "Hello";
     }
 
+    @AfterEach
+    void cleanUp() {
```
casid: This adds only noise to the test.

`output` is a non-static field and as such will be recreated for each unit test method anyways.

jte-runtime/src/main/java/gg/jte/runtime/Template.java
casid: Good point. If the test doesn't actually test it, I would remove it.

I think it's okay to then just add the `throws e;` here. Just in case to make sure that we don't swallow a potential exception by accident.

casid (changes requested):
Thanks for the investigation! 👍 

Just a little clean-up and then this is good to go.

casid:
Thanks for your submission @iTob191 💯

PR #531, opened 2026-03-20
Add `@javax.annotation.processing.Generated` annotation to generated Java source code files

Petersoj:
Generated code should be annotated with `@javax.annotation.processing.Generated` so that static analysis tools like [Error Prone](https://errorprone.info) can ignore generated code.


edward3h:
The files modified here are just part of the "jte-models" extension, not the primary template engine. Was that your intention?

Petersoj:
@edward3h I updated `jte/src/main/java/gg/jte/compiler/java/JavaCodeGenerator.java` as well, which, if I'm not mistaken, is for the primary template engine.

casid (approved):
LGTM.

edward3h:
@Petersoj sorry I missed that!

edward3h (approved):

PR #532, opened 2026-03-21
Add `staticImplementationSingleton` feature to `jte-models`

Petersoj:
For `jte-models`, the generated `StaticTemplates` class has no instance fields and its methods are effectively functions, so adding the ability to turn it into a singleton instance makes it easier to use e.g. `TEMPLATES.myHtmlTemplate().render()` instead of `new StaticTemplates().myHtmlTemplate().render()`.

### DIFF
--- a/docs/gradle-plugin.md
+++ b/docs/gradle-plugin.md
@@ -143,16 +143,17 @@ See details about it in the [jte-models documentation](jte-models.md).
 
 The following parameters are available for this extension:
 
-| Parameter                         | Description                                                                                                                                                                                                 | Default |
-|-----------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
-| `interfaceAnnotation`             | The FQCN of the annotation to add to the generated interface                                                                                                                                                | None    |
-| `implementationAnnotation`        | The FQCN of the annotation to add to the generated implementation classes                                                                                                                                   | None    |
-| `staticImplementationAnnotation`  | The FQCN of the annotation to add to the generated static implementation classes                                                                                                                            | None    |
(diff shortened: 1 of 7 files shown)


kelunik:
Hey @Petersoj,

Why does this have to be an option? Shouldn't it always be enabled?

Also, if things can be static, shouldn't all methods be static rather than creating an instance?

edward3h:
Nice! Something I had thought about doing but didn't get around to.

edward3h:
> Why does this have to be an option? Shouldn't it always be enabled?

I think it should be optional. The preferred way to use the models is just depend on the Templates interface and let your dependency injection give the implementation.

> 
> Also, if things can be static, shouldn't all methods be static rather than creating an instance?

The methods are implementing a common Templates interface, so can't be static.

edward3h (approved):

edward3h:
@Petersoj you have a NullPointerException in the Gradle tests. You should be able to reproduce it locally if you run `./test/build-local.sh`

Petersoj:
@kelunik @edward3h actually, I think it'd be better to implement a static wrapper class that exposes the singleton instance methods as static methods for the Java `StaticTemplates` class. This would be more consistent with how Kotlin's `object` class works (I think). The methods could be statically imported, which in my opinion looks cleaner. Additionally, I think the option should be enabled by default. See my latest commit (which I didn't make a force-push, so feel free to squash my commits upon merge).

Petersoj:
Guess I should actually run the unit tests locally instead of hoping they pass in GitHub Actions lol. I'll fix the Kotlin implementation shortly.

Petersoj:
Actually, I think using a companion object for the Kotlin singleton is a better approach than using `object`, as using a companion object is more consistent with how Java singletons are implemented and allows the Kotlin `StaticTemplates` class to be instantiated and extended.

edward3h:
Thanks!

PR #541, opened 2026-04-01
Fix usePrecompiledTemplates missing from Spring configuration metadata in starter-4-core

garvit-joshi:
Renamed `usePreCompiledTemplates()` to `isUsePrecompiledTemplates()` in `JteProperties` to follow JavaBean naming conventions. The `spring-boot-configuration-processor` only recognizes boolean getters prefixed with `is`/`get`, `gg.jte.use-precompiled-templates` was silently excluded from the generated metadata, breaking IDE auto-complete.


edward3h:
This might need to keep the original named method as well, to avoid breaking existing users.

garvit-joshi:
Done

edward3h:
Thank you for your contribution!

PR #544, opened 2026-04-07
Ignore csp.withgoogle.com in link validation

### LINKED ISSUE #543: validate-external-links sometimes fails on https://csp.withgoogle.com/docs/strict-csp.html
Seems to be a flaky test. Sometimes the check of https://csp.withgoogle.com/docs/strict-csp.html succeeds, other times it gets a 500 response.

I don't think it's particularly helpful to have a workflow fail on a resource that is outside of the control of this project. We should add this URL to the exclude list.

utafrali:
Fixes #543

The domain keeps timing out in link validation. Easier to just skip it in the ignore file.

### DIFF
--- a/.github/workflows/validate-links.yml
+++ b/.github/workflows/validate-links.yml
@@ -17,7 +17,8 @@ jobs:
             --exclude "@template"
             --exclude test.com
             --exclude maven.apache.org
-            --exclude w3.org'
+            --exclude w3.org
+            --exclude "csp\.withgoogle\.com"'
           jobSummary: true
           format: markdown
           fail: true
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,1 @@
+csp\.withgoogle\.com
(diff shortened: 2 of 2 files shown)


edward3h:
Other domains are as "exclude" arguments in `.github/workflows/validate-links.yml`. Looks like this works differently?

utafrali:
Fair point - want me to add it to the exclude list instead?

edward3h:
I don't know the original intention, but it would make sense to follow the existing setup.

utafrali:
Done, pushed the fix.

edward3h:
> Done, pushed the fix.

It doesn't look like the PR changed. Can you check?

utafrali:
Done, pushed the fix.

utafrali:
Done, pushed the fix.

edward3h:
Thank you for your contribution!

utafrali:
Thank you! Glad the contribution was useful.

PR #550, opened 2026-04-25
Extension to generate NullMarked package-info

### LINKED ISSUE #528: JSpecify null safety annotations on JTE generated classes
Recently, Spring Boot 4 introduced JSpecify annotations on all the Spring classes to improve null safety. 
https://spring.io/blog/2025/11/12/null-safe-applications-with-spring-boot-4 
This is a good direction for Java in general.

I have also started using JSpecify annotations on my own code and I am using NullAway and ErrorProne to report possible null issues.

What I have found is that ErrorProne is failing my builds due to the JTE classes generated from my templates that are not annotated with the @NullMarked annotation. I have a workaround for my issue documented here:
https://github.com/spring-gradle-plugins/nullability-plugin/issues/32

But I do think it would be a good thing if JTE generated classes are also @NullMarked annotated.
It could be helpful to non Spring Boot users of JTE as well.

edward3h:
To help with JSpecify support, #528 
Adds a new extension, jte-nullmarked. When enabled, it will generate a `package-info.java` in each package of generated template code, with the `@NullMarked` annotation.


edward3h:
The "Validate External Links" failure is because this change adds a new documentation page and links to it.

edward3h:
Also I messed up the commit history 👎🏻

jte-nullmarked/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>gg.jte</groupId>
+        <artifactId>jte-parent</artifactId>
+        <version>3.2.4-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jte-nullmarked</artifactId>
```
kelunik: Should this be jte-jspecify-nullmarked?

edward3h: It makes sense to be specific. I'll change it.

test/jte-runtime-cp-test-nullmarked/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+            <plugin>
+                <groupId>gg.jte</groupId>
+                <artifactId>jte-maven-plugin</artifactId>
+                <version>3.2.4-SNAPSHOT</version>
+                <configuration>
+                    <sourceDirectory>${basedir}/src/main/jte</sourceDirectory>
+                    <contentType>Html</contentType>
+                    <extensions>
+                        <extension>
+                            <className>gg.jte.nullmarked.NullMarkedExtension</className>
```
kelunik: Should this be `gg.jte.jspecify.NullMarkedExtension`?

jte-jspecify-nullmarked/src/main/java/gg/jte/jspecify/NullMarkedExtension.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        return templateDescriptions.stream()
+                .map(TemplateDescription::packageName)
+                .distinct()
+                .map(pkg -> writePackageInfo(config.generatedSourcesRoot(), pkg))
+                .collect(Collectors.toList());
+    }
+
+    private Path writePackageInfo(Path sourcesRoot, String packageName) {
```
kelunik, line 37: Should we annotate packages or rather the generated classes? What if you want to progressively upgrade to nullmarked? Should we have annotations on templates similar like annotations on parameters?

edward3h: `@NullMarked` is applied at class, package or module level. While a template is generated into a class, in practice it behaves as a single method. Therefore, a package is the right level. (Since we aren't supporting modules.)

So the question is, do we want a way to include or exclude by package?

If you want more progressive, add `@NonNull` or `@Nullable` to individual params.

casid:
Pretty cool!

I have recently released the jte plugin with the annotation support you have added.

I'd wait with the next jte release for this PR, then we'll have a very nice feature set to release.

test/jte-runtime-cp-test-nullmarked/pom.xml
edward3h: Agree - I changed it.

jte-jspecify-nullmarked/src/main/java/gg/jte/jspecify/NullMarkedExtension.java
edward3h: @kelunik Do you have any more thoughts on this?

kelunik: Let me collect some more feedback internally at work, I'll get back to you soon.

edward3h: @kelunik Did you get any more feedback?

b-strauss: Hi, I have talked with @kelunik about this. It would be nice to have a way to annotate the whole template. We could then migrate templates step by step and, after all templates have been migrated, activate the plugin and remove the annotation again. Having to annotate all params individually feels a bit cumbersome.

edward3h: Thanks for following up.

This PR adds an extension. That is, it doesn't touch the core of JTE. Extensions can create additional files, but can't modify the files generated by the core. Adding `@NullMarked` to individual templates _would_ require changes in the core. It's a different scope of work, so belongs in a different PR.

kelunik: @edward3h yes, we discussed that, too. The migration path can be added as a separate feature later, no need to change the extension for that.

edward3h:
This has been open for a while and had some reasonable discussion, so I'll go ahead and merge.

PR #552, opened 2026-04-30
Test and fix for parameter with annotations using "="

### LINKED ISSUE #433: Annotated parameters
Currently, the parameter string parser messes up completely when you attempt to annotate a parameter type, e.g.
```jte
@param @NotNull String myParam
```
In the generated code, the generator mostly uses the annotation without the type in place of the type, and the type where you would expect the variable name.

Annotated parameters would be really useful to automatically fill and check the parameters, though only if the annotations would actually be exposed by the API. Currently, we only get the raw types with `getParamInfo()`. I would propose just returning the complete `java.lang.reflect.Parameter` object instead of `Parameter.getType()`, maybe in a method `getExtendedParamInfo()`. That would, as a bonus, also allow you to receive generic types properly using `Parameter.getParameterizedType()`.

edward3h:
Adds more test cases covering a variety of forms of annotations on params.

Fix the parser by detecting opening and closing parentheses. Resolves #433

### DIFF
--- a/jte/src/main/java/gg/jte/compiler/java/JavaParamInfo.java
+++ b/jte/src/main/java/gg/jte/compiler/java/JavaParamInfo.java
@@ -12,6 +12,7 @@ public class JavaParamInfo {
         int nameEndIndex = -1;
         int defaultValueStartIndex = -1;
         int genericDepth = 0;
+        int parenDepth = 0;
 
         boolean stringMode = false;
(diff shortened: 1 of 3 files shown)


kelunik (approved):

edward3h:
The related fix for jte-intellij - https://github.com/casid/jte-intellij/pull/63

Issue #488, opened 2025-10-23
How do I compile templates without Maven or Gradle?


peterlindstrom123:
I'm not using either build tool in my Clojure-based project.

How do I use compiled Templates?

peterlindstrom123:
Thank you for the response!

> create a TemplateEngine, configure it with settings such as source and target directories, then call generateAll() or precompileAll().

That sounds fairly straightforward, but does that result in the template class files being generated? Where do they go by default?

It seems I'd need to point a TemplateEngine at precompiled class files, to use precompiled templates.

**EDIT:** Oh.. I guess the class files would go in the "target" directory? 😅

peterlindstrom123:
By the way,

> Roughly speaking, you would need to create a TemplateEngine, configure it with settings such as source and target directories, then call generateAll() or precompileAll().

It would be good to see this stuff mentioned in the documentation. Now it _looks_ like it would be complicated without using the Maven/Gradle plugin, but as far as I understand, using precompiled templates manually should actually be quite simple.

(I haven't gotten through it yet though)

casid:
I think it is not a very common use case to do this.

There is some Javadoc in `gg.jte.TemplateEngine#precompileAll(java.util.List<java.lang.String>)` and `gg.jte.TemplateEngine#generateAll`, but the best reference for it would be to look at the code of `gg.jte.maven.CompilerMojo#execute`

casid:
Yes exactly, all Java projects I know use one of the two build tools.

edward3h:
A way to make a more generally usable build could be to make a CLI. [Picocli](https://picocli.info/) can do the work of parsing arguments, so the logic would end up as a relatively thin wrapper around `TemplateEngine`.

edward3h:
@peterlindstrom123 I put up a PR that adds a CLI app for the code generation https://github.com/casid/jte/pull/560

Do you think this addresses the issue? Or you could use it as an example. The actual implementation is just the `Main` class. The rest is test code and documentation.

PR #560, opened 2026-07-08
CLI app for standalone code generation

edward3h:
Address #488 by adding a CLI to generate code from templates without a build plugin.

To keep it small for an initial version:
* Java (.jte) templates only, not .kte. Otherwise it would need to shade in the Kotlin compiler or figure out adding it to the classpath another way.
* Extensions (e.g. jte-models) are not supported. Also awkward classpath dependencies.

### DIFF
--- a/jte-cli/src/main/java/gg/jte/cli/Main.java
+++ b/jte-cli/src/main/java/gg/jte/cli/Main.java
@@ -1,0 +1,91 @@
+package gg.jte.cli;
+
+import gg.jte.ContentType;
+import gg.jte.TemplateEngine;
+import gg.jte.resolve.DirectoryCodeResolver;
+import gg.jte.runtime.Constants;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.Callable;
+
+@Command(name = "jte-cli", mixinStandardHelpOptions = true,
+        description = "Generates Java sources from jte templates, without needing Maven or Gradle.")
+public class Main implements Callable<Integer> {
+
+    @Option(names = {"-s", "--source-directory"}, required = true, description = "The directory where template files are located.")
+    public Path sourceDirectory;
+
+    @Option(names = {"-t", "--target-directory"}, required = true, description = "Destination directory to store generated sources.")
+    public Path targetDirectory;
+
+    @Option(names = {"-c", "--content-type"}, required = true, description = "The content type of all templates: ${COMPLETION-CANDIDATES}.")
+    public ContentType contentType;
+
+    @Option(names = "--trim-control-structures", description = "Trims control structures, resulting in prettier output.")
+    public boolean trimControlStructures;
+
+    @Option(names = "--html-tags", split = ",", description = "Intercepts the given html tags during template compilation.")
+    public String[] htmlTags;
+
+    @Option(names = "--html-comments-preserved", description = "Preserves HTML/CSS/JS comments, which are omitted by default when using Html content type.")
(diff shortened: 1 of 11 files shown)


jte-cli/src/main/java/gg/jte/cli/Main.java
```diff
@@ -0,0 +1,91 @@
...
+    @Option(names = {"-c", "--content-type"}, required = true, description = "The content type of all templates: ${COMPLETION-CANDIDATES}.")
+    public ContentType contentType;
+
+    @Option(names = "--trim-control-structures", description = "Trims control structures, resulting in prettier output.")
+    public boolean trimControlStructures;
+
+    @Option(names = "--html-tags", split = ",", description = "Intercepts the given html tags during template compilation.")
+    public String[] htmlTags;
+
+    @Option(names = "--html-comments-preserved", description = "Preserves HTML/CSS/JS comments, which are omitted by default when using Html content type.")
```
kelunik: I'd expect this to be `--preserve-html-comments`

jte-cli/src/main/java/gg/jte/cli/Main.java
```diff
@@ -0,0 +1,91 @@
...
+import picocli.CommandLine.Option;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.Callable;
+
+@Command(name = "jte-cli", mixinStandardHelpOptions = true,
+        description = "Generates Java sources from jte templates, without needing Maven or Gradle.")
+public class Main implements Callable<Integer> {
```
kelunik, line 11: Should there be subcommands like `compile` for future extension?

jte-cli/src/main/java/gg/jte/cli/Main.java
```diff
@@ -0,0 +1,91 @@
...
+    @Option(names = "--binary-static-content", description = "UTF-8 encodes all static template parts at compile time.")
+    public boolean binaryStaticContent;
+
+    @Option(names = "--package-name", description = "The package name, where template classes are generated to. Defaults to ${DEFAULT-VALUE}.")
+    public String packageName = Constants.PACKAGE_NAME_PRECOMPILED;
+
+    @Option(names = "--target-resource-directory", description = "Directory in which to generate non-java files (resources).")
+    public Path targetResourceDirectory;
+
+    public static void main(String[] args) {
```
kelunik, line 13: There should likely be a way to specify a class path?

jte-cli/src/main/java/gg/jte/cli/Main.java
edward3h: That makes sense. I changed it.

jte-cli/src/main/java/gg/jte/cli/Main.java
edward3h: I assume that a user will already have a JDK and some other build mechanism that compiles Java, so I don't see a need for a `compile` command specifically. As an initial MVP, I think it would be ok to come back and add subcommands later if we see a need.

jte-cli/src/main/java/gg/jte/cli/Main.java
edward3h: classpath is not necessary for the basic generation - it just uses the text from the templates without necessarily having access to classes.

classpath would be needed for Kotline/kte support, and extensions, but I deliberately excluded those as noted in the PR description.

jte-cli/src/main/java/gg/jte/cli/Main.java
kelunik: My point is adding subcommands later will either be a BC break or complicate things.

edward3h: I separated it into a subcommand as you suggest.

PR #564, opened 2026-07-20
make validate links run against a local build of docs

edward3h:
validate links runs against the published version of the documentation site. This means that links referring to newly added docs cause errors, because they haven't been published yet.

This change makes the workflow build a local version of the docs, and validate against that instead of the published version. It should avoid the problem.


edward3h:
The workflow still failed - investigating...

edward3h:
Workflow passes now 👍🏻

PR #565, opened 2026-07-26
Allow for content type overwrite in spring boot servlet views

### LINKED ISSUE #561: Spring boot integration doesn't allow for modifying content type dynamically
Howdy folks, 

I have this [repository](https://github.com/rferreira/sample-modern-java-web) on modern Java web development using JTE for rendering  and a user pointed out to me that they could not render [turbo streams](https://turbo.hotwired.dev/handbook/streams) - which essentially expects a custom content type of `text/vnd.turbo-streams.html`. Albeit there are work arounds (with a custom View implementation) I think this is something the JTE starter should support natively. 

For example: 

```java

    public record TurboStream(List<Integer> ints) {}
    @GetMapping(value = "/stream", produces = "text/vnd.turbo-streams.html")
    public FragmentsRendering stream(Model model) {

        model.addAttribute("model", new TurboStream(List.of(1, 2, 3)));

        return FragmentsRendering.fragment("stream")
                .build();

    }
```

returns the correct body with the incorrect response type since JTE sets it twice two `text/html;charset=UTF-8` 
* https://github.com/casid/jte/blob/main/jte-spring-boot-starter-4/src/main/java/gg/jte/springframework/boot/autoconfigure/JteViewResolver.java 
* https://github.com/casid/jte/blob/main/jte-spring-boot-starter-4/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java#L31 

Ideally we can make JTE only set the content type if it's not set or another route that still allows Spring to overwrite it when a produces is set

rferreira:
* Removed content type overwrite from JteView allowing for it to be set in the controller and added test pinning the behavior
* On the reactive side, I added a test but the view dispatch flow does not allow for content type to be overwritten, so I used the test to pin the behavior for posterity
* I confirmed this behavior matches what thymeleaf is doing as well.

AI disclosure: I used AI to analyze thymeleaf's codebase and compare their implementation with ours. 

Fixes #561

### DIFF
--- a/jte-spring-boot-starter-4/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
+++ b/jte-spring-boot-starter-4/src/main/java/gg/jte/springframework/boot/autoconfigure/JteView.java
@@ -5,7 +5,6 @@ import gg.jte.output.PrintWriterOutput;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import org.jspecify.annotations.NonNull;
-import org.springframework.http.MediaType;
 import org.springframework.web.servlet.view.AbstractTemplateView;
 
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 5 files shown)


rferreira:
I think `validate external links` was broken before my PR as I didn't touch any links.

edward3h:
> I think `validate external links` was broken before my PR as I didn't touch any links.

Correct. I have a different fix up for that. I'll go ahead and merge it.

edward3h (approved):

rferreira:
thanks @edward3h

rferreira:
@edward3h can we get this merged?

edward3h:
Yes. I allowed some time in case anyone else wanted to comment. Then forgot...

